PR #8, opened 2015-11-26
Patch

retokromer:


ablwr:
This should be merged into gh-pages, not master!

retokromer:
Now it should be correct.

PR #11, opened 2015-11-27
now should be better

retokromer:


ablwr:
Wow Reto this is great!!! So clean!

ablwr:
I might have to fiddle with the CSS to get the bold to be more bold.

PR #23, opened 2015-12-21
More consistent -c copy explanations

kieranjol:
Vast majority of explanations were "use stream copy mode to re-mux instead of re-encode". A few variations existed for some explanations.


retokromer:
-c copy on the one hand, and -c:v copy or -c:a copy on the other are **NOT** the same!

retokromer:
Only -c copy use stream copy mode to re-mux instead of re-encode, not the others.

dericed:
`-c copy` is to copy all mapped streams
`-c:a copy` is to copy all mapped audio streams
`-c:v copy` is to copy all mapped video streams

kieranjol:
Apologies, I assumed that the generic explanation for `-c copy` and `-c:a copy` was sufficient. PR closed.

retokromer:
Hello @kieranjol, I think a more consistent explanation is indeed fine! There is a speed difference if it is re-recoded or "only" re-muxed.

PR #24, opened 2015-12-24
add text watermark

kieranjol:
Hopefully the -vf description doesn't break any style constraints. Haven't properly used html since geocities.
From what I can tell, fontsize only accepts an integer, so any relative value must be generated via scripts. Please let me know if there's another way, similar to the x/y placement.

### DIFF
--- a/index.html
+++ b/index.html
@@ -601,13 +601,13 @@ path_name_and_extension_to_the_last_file</i></pre></dd>
           <dt>-vf "drawtext=fontfile=<i>font_path</i>:fontsize=<i>font_size</i>:text=<i>watermark_text</i>:fontcolor=<i>font_colour</i>:alpha=0.4:x=w/2-tw/2:y=h/2-th/2"</dt><dd>This calls the drawtext filters with the following options:
           <dl>
             <dt>fontfile=<i>font_path </dt><dd> Set path to font. For example in OSX: <code>"fontfile=/Library/Fonts/AppleGothic.ttf"</code></dd>
-            <dt>fontsize=<i>font_size </dt><dd> Set font size. <code>24</code> is a good value for SD. Ideally this is proportional to video size, for example use ffprobe to acquire video height and divide by 12</dd>
+            <dt>fontsize=<i>font_size </dt><dd> Set font size. <code>24</code> is a good value for SD. Ideally this value is proportional to video size, for example use ffprobe to acquire video height and divide by 14.</dd>
             <dt>text=<i>watermark_text</i> </dt><dd> Set the content of your watermark text. For example: <code>"text="FFMPROVISR HACK DAY NEVAR ENDS"</code></dd>
             <dt>fontcolor=<i>font_colour</i> </dt><dd> Set colour of font. Can be a text string such as <code>"fontcolor=white"</code> or a hexadecimal value such as <code>"fontcolor=0xFFFFFF"</code></dd>
             <dt>alpha=0.4</dt><dd> Set transparency value.</dd>
             <dt>x=w/2-tw/2:y=h/2-th/2</dt><dd> Sets <i>x</i> and <i>y</i> coordinates for the watermark. These relative values will centre your watermark regardless of video dimensions.</dd>
           </dl>
-          <dt><i>output_file</i></dt><dd>path, name and extension of the output file</dd>
+          <dt><i>output_file</i></dt><dd>path, name and extension of the output file.</dd>
         </dl>
       </div>
     </div>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -586,6 +586,34 @@
...
+        <h3>Create centered, transparent text watermark</h3>
+        <p>E.g For creating access copies with your institutions name</p>
+        <p><code>ffmpeg -i <i>input_file</i> -vf drawtext="fontfile=<i>font_path</i>:fontsize=<i>font_size</i>:text=<i>watermark_text</i>:fontcolor=<i>font_colour</i>:alpha=0.4:x=w/2-tw/2:y=h/2-th/2" <i>output_file</i></code></p>
+        <dl>
+          <dt>ffmpeg</dt><dd>starts the command</dd>
+          <dt>-i <i>input_file</i></dt><dd>path, name and extension of the input file</dd>
+          <dt>-vf drawtext=</dt><dd>This calls the drawtext filters with the following options:
+          <dl>
+            <dt>fontfile=<i>font_path</i></dt><dd> Set path to font. For example in OSX: <code>"fontfile=/Library/Fonts/AppleGothic.ttf"</code></dd>
+            <dt>fontsize=<i>font_size</i></dt><dd> Set font size. <code>35</code> is a good starting point for SD. Ideally this value is proportional to video size, for example use ffprobe to acquire video height and divide by 14.</dd>
```
ablwr, line 604: Does this number represent max pixel height?

index.html
```diff
@@ -586,6 +586,34 @@
...
+            <dt>text=<i>watermark_text</i> </dt><dd> Set the content of your watermark text. For example: <code>"text="FFMPROVISR HACK DAY NEVAR ENDS"</code></dd>
```
ablwr: I think there's an extra " before text here. Or maybe one is missing? I'm looking at line 606 too where fontcolor=color is wrapped in quotes.

index.html
kieranjol, line 604: I'm not certain, the docs are vague: 
https://ffmpeg.org/ffmpeg-filters.html#Syntax

```
The font size to be used for drawing text. The default value of fontsize is 16.
```

index.html
kieranjol: My bad, I'll remove the quotes.

index.html
```diff
@@ -586,6 +586,34 @@
...
+            <dt>fontcolor=<i>font_colour</i> </dt><dd> Set colour of font. Can be a text string such as <code>"fontcolor=white"</code> or a hexadecimal value such as <code>"fontcolor=0xFFFFFF"</code></dd>
+            <dt>alpha=0.4</dt><dd> Set transparency value.</dd>
+            <dt>x=w/2-tw/2:y=h/2-th/2</dt><dd> Sets <i>x</i> and <i>y</i> coordinates for the watermark. These relative values will centre your watermark regardless of video dimensions.</dd>
```
kieranjol: Gonna change this to the example used in the ffmpeg docs `x=(w-text_w)/2:y=(h-text_h)/2`
https://ffmpeg.org/ffmpeg-filters.html#Examples-39

ablwr:
I just pulled your branch and looked! :+1: LGTM :+1:

kieranjol:
Phew! Thanks for the feedback, Ashley.

ablwr:
Hope I didn't scare you. :smiling_imp:

kieranjol:
Nah, just a bit embarrassed :] P.S Happy Holidays!

retokromer:
In other recipes we used mainly -filter:v instead of -vf, therefore I tried to have only one syntax. I added here a remark that’s the same. In fact, for many parameters, there are three different syntaxes… (Perhaps we should add everywhere at the end a short note with the alternative syntaxes?)

kieranjol:
I think it's important to have consistent terminology across the recipes. Maybe it's easier to just use one term so that an explanatory note is not required? I wonder if -vf is more commonly used than -filter:v? The ffmpeg filter documentation uses -vf: https://ffmpeg.org/ffmpeg-filters.html
Edit: The main ffmpeg doc says that `-vf` is an alias for `-filter:v`. Perhaps it's more self descriptive to use `filter:v`, and there is more consistency with terms like `filter_complex` than when using `-vf`.

ablwr:
@retokromer and I talked about this before and settled on -filter:v for the reasons you mention (more descriptive, -vf is an alias of).

kieranjol:
Awesome, I'll fix the -vf in my contributions later, and if it's ok with @retokromer , remove the explanatory note on -vf.

dericed:
+1 for -vf.

retokromer:
For me it’s indifferent for _ffmprovisr_. (I prefer -filter:v for my own teaching.) Please let me know what you prefer: I can quickly change every occurrence with one command…

PR #26, opened 2015-12-25
add burn in timecode

kieranjol:


index.html
```diff
@@ -614,6 +614,35 @@
...
 </div>
 <!-- ends Text watermark -->
+<!-- Burn in timecode-->
+<span data-toggle="modal" data-target=".burn_in_timecode"><button type="button" class="btn btn-default" data-toggle="tooltip" data-placement="bottom" title="Burn in timecode ">Burn in timecode</button></span>
+<div class="modal fade burn_in_timecode" tabindex="-1" role="dialog" aria-labelledby="myLargeModalLabel">
+  <div class="modal-dialog modal-lg">
+    <div class="modal-content">
+      <div class="well">
+        <h3>Create a burnt in timecode on your image</h3>
+        <p><code>ffmpeg -i <i>input_file</i> -vf drawtext="fontfile=<i>font_path</i>:fontsize=<i>font_size</i>:timecode=<i>starting_timecode</i>:fontcolor=<i>font_colour</i>:box=1 :boxcolor=<i>box_colour</i>:rate=<i>timecode_rate</i>:x=(w-text_w)/2:y=h/1.2" <i>output_file</i></code></p>
```
kieranjol, line 624: I added a space before `box=1` as the formatting broke in Ubuntu Firefox and windows 7 Chrome without it. As the drawtext filter is wrapped in quotes, spacing doesn't matter in terms of getting a working ffmpeg command, but currently my ffmprovisr code has inconsistent spacing and I don't know how to fix. Here's what it looks like:
screenshot from 2015-12-25 01 13 00

retokromer:
I cannot check this at the moment, but the information in my brain is that, for an obscure reason, the timecode in FFmpeg is not 00:00:00[.,:;]NN but MMM with milliseconds and not frames at the end. I can check that probably tomorrow (without quoting marks ;-)
Thank you very much, @kieranjol, for your contributions!

kieranjol:
My experience has always been frames rather than milliseconds.
The ffmpeg filter docs hh:mm:ss[:;.]ff when describing the drawtext timecode, but milliseconds is surely possible too,I've just never used it.

PR #36, opened 2016-01-09
adds section for the different kinds of test videos that can be gener…

ablwr:
…ated on the fly

### DIFF
--- a/index.html
+++ b/index.html
@@ -553,6 +553,74 @@
   <!-- ends Check FFV1 Fixity  -->
 
   </div>
+
+  <div class="well">
+    <h4>Test videos</h4>
+
+  <!-- Mandelbrot -->
+  <span data-toggle="modal" data-target=".mandelbrot"><button type="button" class="btn btn-default" data-toggle="tooltip" data-placement="bottom" title="Make a mandelbrot test pattern video">Mandelbrot</button></span>
+  <div class="modal fade mandelbrot" tabindex="-1" role="dialog" aria-labelledby="myLargeModalLabel">
+    <div class="modal-dialog modal-lg">
+      <div class="modal-content">
+        <div class="well">
+          <h3>Makes a mandelbrot test pattern video</h3>
+          <p><code>ffmpeg -f lavfi -i mandelbrot -t 10 <i>output_file</i></code></p>
+          <dl>
+            <dt>ffmpeg</dt><dd>starts the command</dd>
+            <dt>-f lavfi</dt><dd>tells ffmpeg to use the Libavfilter input virtual device <a href="http://ffmpeg.org/ffmpeg-devices.html#lavfi">[more]</a></dd>
+            <dt>-i mandelbrot</dt><dd>asks for the madelbrot test filter as input</dd>
+            <dt>-t 10</dt><dd>specifies recording time of 10 seconds</dd>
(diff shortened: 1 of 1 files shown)


kieranjol:
I think this is a brilliant addition, I like the extension /container explanation. Can't verify this ATM but ,iirc, do these tests create Rgb video by default? Could be nice to add a video codec and pixel format option so people know they can generate the test files in whatever format they want, similar to what you've written about container, something like `-c:v libx264 -pix_fmt yuv420p`?

ablwr:
:+1: @kieranjol ! I believe you are so right and that's such a good idea.

retokromer:
I did some tests on my website. I think SD and HD bars with `ffplay` could also be useful. (At least I use this to test the settings, before presentations.)
http://avpres.net/FFmpeg/#ch7

retokromer:
I second @kieranjol about adding format.

ablwr:
I'll do my favorite thing -- create an issue and assign myself.

ablwr:
:+1: about your comment on ffplay, @retokromer

retokromer:
Hmm… isn’t it quicker to change
directly the code? ;-)

retokromer:
BTW: I think this weekend I can also add some `ffprobe` stuff we use on both ffmprovisr and website.

ablwr:
;) helps me keep my github streak going ;)

retokromer:
@ablwr @kieranjol I suggest to add the `target="_blank"` parameter to the `<a>` tag for external links. If it's OK with you I will made it.

ablwr:
@retokromer Good idea!

retokromer:
– I added the «missing» `target="_blank"` and `ffplay` SMPTE bars for HD and VGA.
– Examples of `ffprobe` will follow. I’m still trying to find the most interesting ones in an archival context.

kieranjol:
I use the following to get starting timecode via ffprobe
`ffprobe -v error -select_streams v:0 -show_entries format_tags=timecode:stream_tags=timecode -of default=noprint_wrappers=1:nokey=1 inputfile`
The reason I have both `format` and `stream` is that some MXFs (some AS-11 test files I have) won't reveal their timecodes if you just look purely at the streams. Not sure how common it is, though. This has generally been reliable for me, as long as timecode exists in some form within a file.
Once I store it as a variable, I can use it in combination with the burnt in timecode command.
I can add this one at some stage unless it seems wonky to ye.

kieranjol:
Also, for access work, I generally find frame size, duration, framerate to be useful in conjunction with other scripts.

retokromer:
Thank you, @kieranjol, your contribution is very valuable, as usual!
I am at the moment testing some «silent» variants, that allow to process the output directly (aka very easily) on other programmes, without headers. Trying to figure out what could be the most useful.

kieranjol:
You're welcome Reto. 
If I understand your meaning correctly, my timecode example is "silent", in that it only produces the value within the terminal, nothing else is shown. Is that what you meant?  Although when actually using it as part of a larger script, I just assign it as a variable and it never appears in the terminal. I'll await your examples and structure my contributions based on yours, because mine are most likely not best practise.

retokromer:
Sorry, as «silent» I was referring to parameters like `-hide_banner` and `-loglevel 0` for my `ffprobe` parts. Not yours.

retokromer:
Hmm… you said extracting «frame size, duration, framerate» etc.? **That** could be an excellent example of use of a bash script! Let’s try to include ffprobe in a bash loop.

PR #45, opened 2016-03-26
add setfield and aspect ratio options

kieranjol:
I think these are the appropriate sections for each recipe?


retokromer:
@kieranjol Should we say which kind of ProRes we use (in Change field order of an interlaced video)?

PR #48, opened 2016-03-28
adds CC-BY license to readme

ablwr:


retokromer:
@ablwr Should CC BY also be mentioned on the left side of index.html?

PR #50, opened 2016-05-06
index.html -H264- Specify yuv420p earlier for h264

kieranjol:
Hey, this might be nitpicking, but I've found that a few people I know have used the first H264 command, which doesn't specify yuv420p, and have had playback issues as a result. Unfortunately, it seems like they didn't read further down to see the clarification re -pix_fmt. As the goal was to send files to people who were using Quicktime/WMP, the files could not be played back by the researchers. Also some of the computers that they were testing on had quicktime or WMP set as a default player, so they even had issues themselves.

While I actually prefer the original recipe, I think switching around the order could make the recipe more useful for newcomers to ffmpeg who don't have VLC or similar players installed.

### DIFF
--- a/index.html
+++ b/index.html
@@ -101,25 +101,21 @@
       <div class="modal-content">
         <div class="well">
           <h3>Transcode to H.264</h3>
-          <p><code>ffmpeg -i <i>input_file</i> -c:v <i>libx264</i> -c:a copy <i>output_file</i></code></p>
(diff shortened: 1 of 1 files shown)


retokromer:
👍 Thank you, @kieranjol!

kieranjol:
:+1:

ablwr:
Yes, thanks always Kieran!

PR #54, opened 2016-05-06
create video from image and audio files

edsu:
If this recipe or something similar is already present please feel free to delete this pull request.

I was trying to convince YouTube to transcribe my audio file, but you can't upload audio to YouTube. _But_ you can take an image file (of a cat of course, but not just cats) and combine it with the audio file to create a video that YouTube will then accept.

I guess it could be useful in other situations where you have an audio file you would like to make available in a video presentation. If this is too much of a corner case feel free to ignore it. But, it was useful to me :-)

### DIFF
--- a/index.html
+++ b/index.html
@@ -950,6 +950,47 @@
     </div>
   </div>
   <!-- ends Images to video -->
+
+  <!-- Create video from image and audio  -->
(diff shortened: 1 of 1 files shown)


ablwr:
I think this is VERY useful!

PR #55, opened 2016-05-07
Comments on style

retokromer:
- If we do **not** code CR and/or LF into the <p> tag, then is works fine more universally (e.g. on my Braille terminal).
- -vf is an alias for -filter:v. If we use it here, then we should at least mention this alias also in the other recipes.
- Here we use the alias -acodec, otherwise the abbreviation -c:a of -codec:a. This could be confusing for newcomers. At least a remark is needed.


ablwr:
Very great!

PR #56, opened 2016-05-10
add trim video option

kfrn:
Although there are a few variations on how one can trim a video, I'd like to suggest this as a useful all-purpose command.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1008,7 +1008,7 @@
       <div class="modal-content">
         <div class="well">
           <h3>Trim a video without re-encoding</h3>
-          <p><code>ffmpeg -i <i>input_file</i> -ss 00:02:00 to 00:55:00 -c copy <i>output_file</i></code></p>
+          <p><code>ffmpeg -i <i>input_file</i> -ss 00:02:00 -to 00:55:00 -c copy <i>output_file</i></code></p>
           <p>This command trims the beginning and end of a video file.</p>
           <dl>
            <dt>ffmpeg</dt><dd>starts the command</dd>
--- a/index.html
+++ b/index.html
@@ -1007,7 +1007,7 @@
     <div class="modal-dialog modal-lg">
       <div class="modal-content">
         <div class="well">
-          <h3>Trim a video with re-encoding</h3>
+          <h3>Trim a video without re-encoding</h3>
           <p><code>ffmpeg -i <i>input_file</i> -ss 00:02:00 to 00:55:00 -c copy <i>output_file</i></code></p>
           <p>This command trims the beginning and end of a video file.</p>
           <dl>
@@ -1017,7 +1017,7 @@
 		   <dt>to 00:55:00</dt><dd>sets out point at 00:55:00</dd>
            <dt>-c copy</dt><dd>use stream copy mode (no re-encoding)</dd>
            <dt><i>output file</i></dt><dd>path, name and extension of the output file</dd>
-		   Note: watch out when using <code>-ss</code> with <code>-c copy</code> when the source is an interframe codec (e.g., H.264). Since ffmpeg must split on i-frames, it will seek to the nearest i-frame to begin the stream copy. 
+		   Note: watch out when using <code>-ss</code> with <code>-c copy</code> when the source is encoded with an interframe codec (e.g., H.264). Since ffmpeg must split on i-frames, it will seek to the nearest i-frame to begin the stream copy. 
 		  </dl>
         </div>
       </div>
--- a/index.html
+++ b/index.html
@@ -370,9 +370,9 @@
     <div class="modal-dialog modal-lg">
       <div class="modal-content">
         <div class="well">
-          <h3>Trim</h3>
+          <h3>Excerpt from beginning</h3>
           <p><code>ffmpeg -i <i>input_file</i> -t <i>5</i> -c copy <i>output_file</i></code></p>
-          <p>This command trims the beginning and end of a video file without re-encoding</p>
+          <p>This command captures a certain portion of a video file, starting from the beginning and continuing for the amount of time (in seconds) specified in the script. This can be used to create a preview file, or to remove unwanted content from the end of the file. To be more specific, use timecode, such as 00:00:05.</p>
           <dl>
             <dt>ffmpeg</dt><dd>starts the command</dd>
             <dt>-i <i>input_file</i></dt><dd>path, name and extension of the input file</dd>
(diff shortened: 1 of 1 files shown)


kieranjol:
Hi!

I think this would serve as a replacement for 'Excerpt from middle' ? https://github.com/amiaopensource/ffmprovisr/blob/gh-pages/index.html#L389

A few points: I believe that `-to` requires a dash for ffmpeg to incorporate it. I'm not sure if `-to` is preferable to `-t`, but I'd say `-to` is fine : https://ffmpeg.org/ffmpeg.html#Main-options

Also according to this: https://trac.ffmpeg.org/wiki/Seeking#Cuttingsmallsections
It could be preferable to have the first `-ss` as an input option, as it might be faster? I could be wrong on this so advice welcome!

Edit: It looks like `-copyts` should be added in order to use `-ss` as an input option if you want to use the timestamps of your input file.

kfrn:
Hi Kieran, you're quite right - how did I miss that command? I guess I was looking for the word 'trim'. Apologies.
And yes, that's a typo - it should be `-to`, I'll correct that.
I prefer using `-to` rather than `-t`, but in many cases you could use either. 
Regarding `-copyts`, I think that's only necessary if you put `-ss` before the input, which indeed is the faster seek option. So it would either be:
`ffmpeg -i input_file -ss 00:02:00 -to 00:55:00 -c copy output_file`  **or**
`ffmpeg -ss 00:02:00 -i input_file -to 00:55:00 -c copy -copyts output_file`
I went with the simpler option, but I welcome discussion on this!
Anyway, since this is a partial duplication of an existing command, what's the best thing to do here - combine, replace, disregard?

kieranjol:
I actually find -to to be really helpful and more human readable. I'd hadn't heard of it until you mentioned it. Perhaps @retokromer could advise on what is an alias of what.
My vote would be for a combination as I think you have some very useful contributions. What do ye think, @ablwr / @retokromer ?

ablwr:
In favor of combining -- or they are diverse enough, no problem in having
multiple similar commands if they teach different things!

kfrn:
Cool, I'll combine in the next day or two. 
May I change the command title to "Trim" or "Trim start and end"?

ablwr:
+1

retokromer:
Personally I prefer «Trim» over «Trim start and end». My taste.

`-t` means a duration in time
`-to` means a specific point in the stream
That is not exactly the same, but often you can use either of it. (There is at least one exception I came across, but I don’t remember the details at the moment.) You **cannot** use both in the same command.

A trim «in the middle» command could be good way to introduce the `-to` parameter. I guess we should replace the existing command by a combination of the two commands, in order to not confuse the newcomers.

kfrn:
Thank you Reto. Yes, and `-t` has precedence. 
I like `-to` since I prefer entering specified timings (and `-ss 00:05:00 -t 10` can easily be expressed as `-ss 00:05:00 -t 00:05:10`, but `-t` doesn't have as much versatility.
I'll try an integrated command and allow you folks to weigh in.

index.html
```diff
@@ -1017,7 +1017,7 @@
 		   <dt>to 00:55:00</dt><dd>sets out point at 00:55:00</dd>
            <dt>-c copy</dt><dd>use stream copy mode (no re-encoding)</dd>
            <dt><i>output file</i></dt><dd>path, name and extension of the output file</dd>
-		   Note: watch out when using <code>-ss</code> with <code>-c copy</code> when the source is an interframe codec (e.g., H.264). Since ffmpeg must split on i-frames, it will seek to the nearest i-frame to begin the stream copy. 
+		   Note: watch out when using <code>-ss</code> with <code>-c copy</code> when the source is encoded with an interframe codec (e.g., H.264). Since ffmpeg must split on i-frames, it will seek to the nearest i-frame to begin the stream copy.
```
retokromer: Please put the Note inside the `<dd>` tag, e.g.
```
<dd>… the output file<br/>
Note: Watch out… </dd>
```
Thank you!

kfrn:
Let me know how it looks, folks :)

index.html
```diff
@@ -370,9 +370,9 @@
     <div class="modal-dialog modal-lg">
       <div class="modal-content">
         <div class="well">
-          <h3>Excerpt from beginning</h3>
+          <h3>Trim</h3>
           <p><code>ffmpeg -i <i>input_file</i> -t <i>5</i> -c copy <i>output_file</i></code></p>
-          <p>This command captures a certain portion of a video file, starting from the beginning and continuing for the amount of time (in seconds) specified in the script. This can be used to create a preview file, or to remove unwanted content from the end of the file. To be more specific, use timecode, such as 00:00:05.</p>
+          <p>This command trims the beginning and end of a video file without re-encoding</p>
```
retokromer: If I remember carefully, the idea of this receipt was e.g. to capture the opening titles for cataloguing purposes.

index.html
```diff
@@ -386,28 +386,33 @@
...
+            <dt>-ss 00:02:00</dt><dd>sets in point at 00:02:00</dd>
+		    <dt>-to 00:55:00</dt><dd>sets out point at 00:55:00</dd>
+            <dt>-c copy</dt><dd>use stream copy mode (no re-encoding)</dd>
+			<dd><i>Note:</i> watch out when using <code>-ss</code> with <code>-c copy</code> if the source is encoded with an interframe codec (e.g., H.264). Since ffmpeg must split on i-frames, it will seek to the nearest i-frame to begin the stream copy.</dd>
             <dt><i>output_file</i></dt><dd>path, name and extension of the output file</dd>
           </dl>
+		  <p>Variation: trim video by setting duration, by using <code>-t</code> instead of <code>-to</code></p>
+		  <p><code>ffmpeg -i <i>input_file</i> -ss 00:05:00 -t 10 -c copy <i>output_file</i></code></p>
+          <dt>-ss 00:05:00 -t 10</dt>
+		  <dd>Beginning five minutes into the original video, this command will create a 10-second-long excerpt.</dd>
```
retokromer: The closing `</dl>` should be placed after this

```
<dt>…</dt>
<dd>…</dd>
```

statement.

index.html
```diff
@@ -386,28 +386,33 @@
...
           <dl>
             <dt>ffmpeg</dt><dd>starts the command</dd>
             <dt>-i <i>input_file</i></dt><dd>path, name and extension of the input file</dd>
-            <dt>-ss <i>5</i></dt><dd>Tells ffmpeg what timecode in the file to look for to start copying, and specifies the number of seconds into the video that ffmpeg should start copying. To be more specific, you can use timecode such as 00:00:05.</dd>
-            <dt>-t <i>10</i></dt><dd>Tells ffmpeg to stop copying from the input file after a certain time, and specifies the number of seconds after which to stop copying. In this case, 10 seconds is specified.</dd>
-            <dt>-c copy</dt><dd>use stream copy mode to re-mux instead of re-encode</dd>
+            <dt>-ss 00:02:00</dt><dd>sets in point at 00:02:00</dd>
+		    <dt>-to 00:55:00</dt><dd>sets out point at 00:55:00</dd>
+            <dt>-c copy</dt><dd>use stream copy mode (no re-encoding)</dd>
+			<dd><i>Note:</i> watch out when using <code>-ss</code> with <code>-c copy</code> if the source is encoded with an interframe codec (e.g., H.264). Since ffmpeg must split on i-frames, it will seek to the nearest i-frame to begin the stream copy.</dd>
```
retokromer: A double `<dd>…</dd>` is syntactically valid HTML, but may cause troubles when «displayed» on interfaces for impaired people. Therefore I personally recommend to use only `dt` and `dd` pairs.

retokromer:
Thank you very much, @kfrn!

The «Trim» section LGTM – with a little HTML cosmetic.

For the «Excerpt from beginning» section: could one English speaker please check it. I understand this differently, but possibly it’s just me. Sorry.

index.html
kfrn: Apologies, I edited this in error. Will reinstate.

index.html
kfrn: I have wrapped the variation code in its own `<dl>` pair, I hope that is permissable.

kieranjol:
As for «Excerpt from beginning» I understood it to mean that the beginning was untouched as well. By the way, I referenced this trim command yesterday for work. Very handy, thanks for the contribution.

kfrn:
Re: the «excerpt» commands, I would probably have expressed them as "trim end of video", etc, as per NLE terminology.

retokromer:
Thank you @kfrn and @kieranjol for your collaboration: most appreciated!

kfrn:
Thank you all for being so welcoming! :-)

index.html
retokromer: Of course it is! :-)

PR #63, opened 2016-05-18
amend concat command

kfrn:
Updating info on .txt input file for concatenation. I've always used the format given on the [ffmpeg wiki](https://trac.ffmpeg.org/wiki/Concatenate); in fact, when I directly followed the current instructions, the command didn't work for me - although that may partly be an HTML formatting issue(?).
Therefore, I'd like to normalize the current ffmprovisr command to the wiki instructions; have also added a note to make things super clear. The file path may be absolute (as shown in example) or relative (e.g. `./cats01.mov`).

### DIFF
--- a/index.html
+++ b/index.html
@@ -702,19 +702,19 @@
           <dl>
             <dt>ffmpeg</dt><dd>starts the command</dd>
             <dt>-f concat</dt><dd>forces ffmpeg to concatenate the files and to keep the same file format</dd>
-            <dt>-i <i>mylist.txt</i></dt><dd>path, name and extension of the input file. This text file contains the list of files to be concatenated and should be formatted as follows:
-            <pre>file '<i>path/to/first_file.ext</i>'
-file '<i>path/to/second_file.ext</i>'
+            <dt>-i <i>mylist.txt</i></dt><dd>path, name and extension of the input file. Per the <a href="https://www.ffmpeg.org/ffmpeg-formats.html#Options">ffmpeg documentation</a>, it is preferable to specify relative rather than absolute file paths, as allowing absolute file paths may pose a security risk.<br>
+            This text file contains the list of files to be concatenated and should be formatted as follows:
+            <pre>file '<i>./first_file.ext</i>'
+file '<i>./second_file.ext</i>'
 . . .
-file '<i>path/to/last_file.ext</i>'</pre>
-In the above, <strong>file</strong> is simply the word "file". Example of an .txt input file:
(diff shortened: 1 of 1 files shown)


kieranjol:
Hi Katherine,

All looks good to me in terms of the commands (I know very little about the html side of things), but I noticed that your Windows path examples are absolute paths. These have been judged 'unsafe' in recent ffmpeg builds:
info here:
https://ffmpeg.org/pipermail/ffmpeg-user/2016-March/031487.html

and
https://ffmpeg.org/pipermail/ffmpeg-user/2016-March/031489.html

and 
https://www.ffmpeg.org/ffmpeg-formats.html#Options

I ended up encountering this issue recently after updating all my builds, so perhaps information about this should be included in the ffmprovisr recipe, with the solution of using `-safe 0` if a user wants to use an absolute path?

Edit: This hasn't been added to the ffmpeg wiki either, so I should try to add to that when I get a chance.

kfrn:
Good to know! I simply went by the ffmpeg wiki on that. I'll update that.

kieranjol:
Not sure exactly when it happened, this commit from March looks like it might be the one: https://github.com/FFmpeg/FFmpeg/commit/689211d5727231c3fe92762d224dbadebdbf4e30

kfrn:
Really good catch. Thanks, Kieran! I've amended, LMK what you reckon.

kieranjol:
LGTM! Thanks for adding that.

index.html
```diff
@@ -702,19 +702,19 @@
           <dl>
             <dt>ffmpeg</dt><dd>starts the command</dd>
             <dt>-f concat</dt><dd>forces ffmpeg to concatenate the files and to keep the same file format</dd>
-            <dt>-i <i>mylist.txt</i></dt><dd>path, name and extension of the input file. This text file contains the list of files to be concatenated and should be formatted as follows:
-            <pre>file '<i>path/to/first_file.ext</i>'
-file '<i>path/to/second_file.ext</i>'
+            <dt>-i <i>mylist.txt</i></dt><dd>path, name and extension of the input file. Per the <a href="https://www.ffmpeg.org/ffmpeg-formats.html#Options">ffmpeg documentation</a>, it is preferable to specify relative rather than absolute file paths, as allowing absolute file paths may pose a security risk.<br>
```
kieranjol, line 705: Awesome :) :+1:

retokromer:
Thank you once more, @kfrn and @kieranjol! BTW: The "old" format works only if the "old" naming rules are followed.

PR #66, opened 2016-06-29
added MKV to MP4

retokromer:
As I am ill, needs to be triple checked.


kieranjol:
Also get well Reto!!

kieranjol:
Can't test but all looks good to me.

PR #74, opened 2016-07-10
'check_framemd5.sh' Bash script for review

retokromer:
- It's more or less the situation as... on February.
- I changed GREEN to BLUE, in order to make the script suitable for users suffering red-green colour blindness.


ablwr:
👍 ⚡

PR #81, opened 2016-07-21
adds idet filter for checking interlacement pattern status

ablwr:
As referenced and described in @dericed's http://www.tate.org.uk/research/publications/sustaining-consistent-video-presentation

### DIFF
--- a/index.html
+++ b/index.html
@@ -641,12 +641,12 @@
       <div class="modal-content">
         <div class="well">
           <h3>Check video file interlacement patterns</h3>
-          <p><code>ffmpeg -i <i>input file</i> -vf idet -f null -</code></p>
+          <p><code>ffmpeg -i <i>input file</i> -filter:v idet -f null -</code></p>
           <p></p>
           <dl>
             <dt>ffmpeg</dt><dd>starts the command</dd>
             <dt>-i <i>input_file</i></dt><dd>path, name and extension of the input file</dd>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -634,6 +634,29 @@
...
+    <div class="modal-dialog modal-lg">
+      <div class="modal-content">
+        <div class="well">
+          <h3>Check video file interlacement patterns</h3>
+          <p><code>ffmpeg -i <i>input file</i> -vf idet -f null -</code></p>
+          <p></p>
+          <dl>
+            <dt>ffmpeg</dt><dd>starts the command</dd>
+            <dt>-i <i>input_file</i></dt><dd>path, name and extension of the input file</dd>
+            <dt>-vf idet</dt><dd>This calls the <a href="http://www.tate.org.uk/research/publications/sustaining-consistent-video-presentation">idet (detect video interlacing type) filter</a>.</dd>
```
retokromer: I guess in the other recipes we used `-filter:v`. Therefore we should here either do the same or at least indicate that `-vf` is an alias.

ablwr: You are right. I updated!

retokromer:
LGTM & merged. Thank you!

index.html
retokromer: I know, I am a pedantic, but it facilitates very much the scripting, when syntax is used consistently.

ablwr: No, I agree, especially for beginners it is good to see consistency in a verbose manner.

PR #83, opened 2016-07-22
add notes about setting properties during stream copy

dericed:

PR #84, opened 2016-07-22
Set stream properties

ablwr:
Don't have permissions to write to @dericed branch directly but here is the HTML updates required to merge, superseding #83

### DIFF
--- a/index.html
+++ b/index.html
@@ -1075,6 +1075,48 @@ e.g.: <code>ffmpeg -f concat -safe 0 -i mylist.txt -c copy <i>output_file</i></c
   </div>
   <!-- ends set field order -->
   
+  <!-- Make stream properties explicate -->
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -271,7 +271,7 @@
           <p><code>ffmpeg -i <i>input_file</i> -c:v copy -aspect 4:3 <i>output_file</i></code></p>
           <dl>
             <dt>ffmpeg</dt><dd>starts the command</dd>
-            <dt>-i <i>input_file</i></dt><dd>path, name and extension of the input file<br/>
+            <dt>-i <i>input_file</i></dt><dd>path, name and extension of the input file</dd>
             <dt>-c:v copy</dt><dd>Copy all mapped video streams.</dd>
```
ablwr, line 275: Found some other rogue break-tags.

retokromer:
I have the impression you got all (cf. https://github.com/amiaopensource/ffmprovisr/pull/83).

PR #88, opened 2016-07-23
revision

retokromer:
- added some { }
- added test diff only if Cygwin
- integrated version in help
- added please wait message
- updated version

Please check it, @kieranjol (I don’t have access to Windows now ;-)

### DIFF
--- a/check_framemd5.sh
+++ b/check_framemd5.sh
@@ -59,7 +59,7 @@ elif [ "${#}" -eq 2 ]; then
     else
         unset md5_tmp
         if [[ $OSTYPE == "cygwin" ]]; then
-            md5_tmp=""${USERPROFILE}/$(basename $2).tmp""
+            md5_tmp=""${USERPROFILE}/$(basename ${2}).tmp""
         else 
             md5_tmp="${HOME}/$(basename ${2}).tmp"
         fi
--- a/check_framemd5.sh
+++ b/check_framemd5.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
-  SCRIPT=$(basename $0)
- VERSION='2016-07-10'
+  SCRIPT=$(basename "${0}")
+ VERSION='2016-07-23'
   AUTHOR='ffmprovisr'
      RED='\033[1;31m'
    BLUE='\033[1;34m'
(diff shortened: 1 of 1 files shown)


ablwr:
Deferring to a Windows user to test and approve.

retokromer:
@kieranjol I’m also wondering if `getopt` is installed by default with Cygwin, or if it can be installed like `diff`. That would allow to greatly improve the input reading and handling.

(BTW: The `getopt` coming on Apple is shitty, and `gnu-getops` needs to be installed instead with `brew`.)

kieranjol:
Hi Reto, I wasn't able to dig out my windows computer, so I'll try to test this out when back at work tomorrow. I've found that the bash terminal that comes with git-for-windows to work much easier 'out of the box' than cygwin, and I generally use that when running bash. I think it's just mingw. Would it be ok to just test on that instead, or would cygwin definitely need to be tested?

retokromer:
Hello Kieran! No, for me it’s completely indifferent. If the Bash from git-for-windows works, then it‘s very fine for me. How do we check if it’s installed?

For something different, in the «bilingual» Linux/Mac context, I do:

``` bash
if [[ ${OSTYPE} == "darwin"* ]]; then
    if ! tmp_loc="$(type -p "$(brew --prefix gnu-getopt)/bin/getopt")" || [ -z "$tmp_loc" ]; then
        brew install gnu-getopt
        brew link --force gnu-getopt
    fi
fi
```

The same way we could check if that Bash terminal is installed under Windows and, if not, to install it. This would made the script more useful, I guess.

retokromer:
Hello @kieranjol, do you know how I can test under Windows if Bash is installed, and if not to install it? This way I could merge and close this pull request.

kieranjol:
Hi Reto,

Firstly the script works perfectly in windows 10 using git bash:

```
kieranjol@MINGW32 ~
$ ./fmd5.sh webcam.mkv out.md5
Please wait...
OK 'webcam.mkv' matches 'out.md5'.

kieranjol@MINGW32 ~
$ ./fmd5.sh webcam.mkv out.md5
Please wait...
ERROR: The following differences were detected between 'webcam.mkv' and 'out.md5':
15c15
< 0,         16,         16,        1,   614400, 2a35f957cdf8d7fd9c8b2e3262241a3f
---
> 0,         16,         16,        1,   614400, 2a35f957cdf8d7fd9c8b2e326224133f

```

 When I use git bash (https://git-for-windows.github.io/) - I don't run it in either of the two windows terminals (cmd and powershell). It uses it's own mingw terminal. So I guess if you're running that, then you know that you have bash installed. I think cygwin might be somewhat similar, although perhaps there's a way to just run it in a standard windows terminal? 

From my experience, git bash just works really well and it's a really simple way to get bash on your windows machine. Unfortunately, my new cheap laptop has a 32-bit version of windows 10 so i can't try our the official bash terminal :[

As for `getopt` - it comes installed with git bash, as does diff:

```
 MINGW32 ~
$ getopt -V
getopt from util-linux 2.26.2

kieranjol@MINGW32 ~
$ diff -v
diff (GNU diffutils) 3.3
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Paul Eggert, Mike Haertel, David Hayes,
Richard Stallman, and Len Tower.
```

PR #95, opened 2016-08-29
fix play image sequence

kieranjol:


kieranjol:
The issue is that the recipe does not currently open due to misnamed `data-target`

retokromer:
Good catch, thanks!

Issue #42, opened 2016-02-11
Integrate this head switching noise cover script


ablwr:
FFmpeg dropped some knowledge into my timeline when I was complaining about the weather and I need to drop that knowledge into this! https://twitter.com/FFmpeg/status/695689188205731846

retokromer:
What kind of filtering would be a good example for an archival context?
https://ffmpeg.org/ffmpeg-filters.html#drawbox

ablwr:
I figured the example they gave was a good one, for covering up the head switching noise. I don't see much of a use for archival purposes but maybe cleaning up something for access?

retokromer:
May this be closed?

ablwr:
Yes!

PR #96, opened 2016-09-01
command for covering head switching noise resolving #42

ablwr:
Hello! This one had been lingering around for quite a while! Free time is great to have!

### DIFF
--- a/index.html
+++ b/index.html
@@ -537,6 +537,36 @@
   </div>
   <!-- ends Create ISO -->
 
+  <!-- Cover head switching noise -->
+  <span data-toggle="modal" data-target="#cover_head"><button type="button" class="btn btn-default" data-toggle="tooltip" data-placement="bottom" title="Cover head switching noise">Cover head</button></span>
+  <div id="cover_head" class="modal fade" tabindex="-1" role="dialog">
+    <div class="modal-dialog modal-lg">
+      <div class="modal-content">
+        <div class="well">
+          <h3>Cover head switching noise</h3>
+          <p><code>ffmpeg -i <i>input_file</i> -vf drawbox=w=iw:h=7:y=ih-h:t=max <i>output_file</i></code></p>
+          <p>This command will draw a black box over a small area of the bottom of the frame, which can be used to cover up head switching noise.</p>
+          <dl>
+            <dt>ffmpeg</dt><dd>starts the command</dd>
+            <dt>-i <i>input_file</i></dt><dd>path, name and extension of the input file</dd>
+            <dt>-vf drawbox=</dt>
+            <dd>This calls the drawtext filter with the following options:
+              <dl>
+                <dt>w=in_w</dt><dd>Width is set to the input width. Shorthand for this command would be w=iw</dd>
+                <dt>h=7</dt><dd>Height is set to 7 pixels.</dd>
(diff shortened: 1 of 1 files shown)


kieranjol:
Looks great to me, and really useful!

dericed:
But why cover it? Why not just crop it off. btw this command, shows the bottom 20 lines with magnification and a grid overlay so it's easier to count how many lines to conceal

```
ffplay some_analogy_vid.mkv -vf crop=iw:20:0:ih-20,scale=iw:ih*20:flags=neighbor,setdar=4/3,drawgrid=width=iw:height=ih/20:t=1:c=yellow
```

ablwr:
If this is somewhat a vanity issue and is for access/playback, wouldn't the crop-off potentially result in finicky video playback -- potentially causing stretch/distortion?

I like this ffplay command though, maybe can integrate into the existing or merits its own script (it's kind of complex to explain out).

dericed:
With cropping the SAR will remain the same, though the DAR will change accordingly, so it shouldn't be stretch by any player that can understand the sar.

index.html
```diff
@@ -537,6 +537,36 @@
...
   <!-- ends Create ISO -->
 
+  <!-- Cover head switching noise -->
+  <span data-toggle="modal" data-target="#cover_head"><button type="button" class="btn btn-default" data-toggle="tooltip" data-placement="bottom" title="Cover head switching noise">Cover head</button></span>
+  <div id="cover_head" class="modal fade" tabindex="-1" role="dialog">
+    <div class="modal-dialog modal-lg">
+      <div class="modal-content">
+        <div class="well">
+          <h3>Cover head switching noise</h3>
+          <p><code>ffmpeg -i <i>input_file</i> -vf drawbox=w=iw:h=7:y=ih-h:t=max <i>output_file</i></code></p>
```
retokromer: In my opinion, it should be said that `-vf` in an alias of `-filter:v`.

index.html
```diff
@@ -537,6 +537,36 @@
...
+          <dl>
+            <dt>ffmpeg</dt><dd>starts the command</dd>
+            <dt>-i <i>input_file</i></dt><dd>path, name and extension of the input file</dd>
+            <dt>-vf drawbox=</dt>
+            <dd>This calls the drawtext filter with the following options:
+              <dl>
+                <dt>w=in_w</dt><dd>Width is set to the input width. Shorthand for this command would be w=iw</dd>
+                <dt>h=7</dt><dd>Height is set to 7 pixels.</dd>
+                <dt>y=ih-h</dt><dd>Y represents the offset, and ih-h sets it to the input height minus the height declared in the previous parameter, setting the box at the bottom of the frame.</dd>
+                <dt>t=max</dt><dd>T represents the thickness of the drawn box. Default is 3. </dd>
```
retokromer: Space before `</dd>` can be deleted.

index.html
```diff
@@ -537,6 +537,36 @@
...
+              </dl>
+            </dd>
+            <dt><i>output_file</i></dt><dd>path and name of the output file.</dd>
```
retokromer: In the other recipes we closed only full sentence with a period. I suggest to cancel it here.

ablwr:
Thanks always, R E T O !

retokromer:
I merge. If the discussion comes to another result we can revert.

ablwr:
Maybe @dericed can add his above script. 👼

PR #100, opened 2016-10-01
housekeeping

retokromer:
- possibly last changes... before using getopt!


retokromer:
Argh! Just noticed that the `getopt` coming with macOS is really shitty. I’ll need to use either:
- `gnu-getopt` (with force linking, or adding an OS checking and adding the path…) or
- `getopts` (which is sadly less powerful).

retokromer:
Please check it! cc: @kieranjol @ablwr 
Thank you very much indeed!

Issue #91, opened 2016-07-25
Script to convert DVD (.iso) to (something else)


ablwr:
@kfrn has agreed to do this. Making an issue so we know it's a work in progress.

retokromer:
Convert to Matroska with all the DVD features, as explained by Steve Lhomme?

kfrn:
I'll dig into this over the next 2 or 3 weeks!
@retokromer, I'm not aware about this, do you have a link?

retokromer:
No, not yet. I am just thinking about Steve Lhomme’s presentation in Berlin on why and how he designed Matroska that way.

retokromer:
@kfrn: I cannot delve deeper into that now, but this may be inspiring: https://github.com/Matroska-Org/dvdmenuxtractor

kfrn:
I didn't forget about this, but my computer is broken at the moment! When it's fixed I'll get onto it

retokromer:
No rush!

ablwr:
Pinging @kfrn !

kfrn:
Hey, thanks for the reminder. :) Sorry folks, life has been crazy in the last while. If it's okay, I'll try to have a good look at this over the next fortnight, and if I don't come up with something decent, it's up for grabs.

retokromer:
Done with https://github.com/amiaopensource/ffmprovisr/pull/103

PR #103, opened 2016-10-30
Add DVD -> H.264 command

kfrn:
For [issue 91](https://github.com/amiaopensource/ffmprovisr/issues/91). This is a relatively simple DVD/ISO ->H.264 transcode, I didn't look at using Matroska to retain menus etc (for future?)

I also made a couple of small edits to the sample template.

### DIFF
--- a/index.html
+++ b/index.html
@@ -373,14 +373,15 @@
         <div class="well">
           <h3>Convert DVD to H.264</h3>
           <p> <code>ffmpeg -i concat:<i>input_file1</i>\|<i>input_file2</i>\|<i>input_file3</i> -c:v libx264 -c:a copy <i>output_file</i>.mp4</code></p>
-          <p>This command allows you to create an H.264 file from a DVD source.</p>
+          <p>This command allows you to create an H.264 file from a DVD source that is not copy-protected.</p>
+          <p>Before encoding, you'll need to establish which of the .VOB files on the DVD or .iso contain the content that you wish to encode. Inside the VIDEO_TS directory, you will see a series of files with names like VTS_01_0.VOB, VTS_01_1.VOB, etc. Some of the .VOB files will contain menus, special features, etc, so locate the ones that contain target content by playing them back in VLC.</p>
           <dl>
             <dt>ffmpeg</dt><dd>starts the command</dd>
             <dt>-i concat:<i>input files</i></dt><dd>lists the input VOB files and directs ffmpeg to concatenate them. Each input file should be separated by a backslash and a pipe, like so:<br>
               <code>-i concat:VTS_01_1.VOB\|VTS_01_2.VOB\|VTS_01_3.VOB</code><br>
               The backslash is simply an escape character for the pipe (<strong>|</strong>).
             <dt>-c:v libx264</dt><dd>sets the codec as H.264</dd>
-            <dt>-c:a copy</dt><dd>audio remains as-is (it is not reencoded)</dd>
+            <dt>-c:a copy</dt><dd>audio remains as-is (no re-encode)</dd>
             <dt><i>output_file.mp4</i></dt><dd>path and name of the output file.</dd>
           </dl>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -365,6 +365,45 @@
...
+  <!-- Rip DVD -->
+  <span data-toggle="modal" data-target="#dvd_to_file"><button type="button" class="btn btn-default" data-toggle="tooltip" data-placement="bottom" title="Basic DVD to file conversion">Convert DVD to H.264</button></span>
+  <div id="dvd_to_file" class="modal fade" tabindex="-1" role="dialog">
+    <div class="modal-dialog modal-lg">
+      <div class="modal-content">
+        <div class="well">
+          <h3>Convert DVD to H.264</h3>
+          <p> <code>ffmpeg -i concat:<i>input_file1</i>\|<i>input_file2</i>\|<i>input_file3</i> -c:v libx264 -c:a copy <i>output_file</i>.mp4</code></p>
+          <p>This command allows you to create an H.264 file from a DVD source that is not copy-protected.</p>
+          <p>Before encoding, you'll need to establish which of the .VOB files on the DVD or .iso contain the content that you wish to encode. Inside the VIDEO_TS directory, you will see a series of files with names like VTS_01_0.VOB, VTS_01_1.VOB, etc. Some of the .VOB files will contain menus, special features, etc, so locate the ones that contain target content by playing them back in VLC.</p>
```
retokromer: In the other recipes typographic ’ are used instead of ' (also below).

index.html
```diff
@@ -365,6 +365,45 @@
...
+          <dl>
+            <dt>ffmpeg</dt><dd>starts the command</dd>
+            <dt>-i concat:<i>input files</i></dt><dd>lists the input VOB files and directs ffmpeg to concatenate them. Each input file should be separated by a backslash and a pipe, like so:<br>
```
retokromer: In HTML 5 it's `<br\>` (also below)

index.html
```diff
@@ -365,6 +365,45 @@
...
+              <code>-i concat:VTS_01_1.VOB\|VTS_01_2.VOB\|VTS_01_3.VOB</code><br>
+              The backslash is simply an escape character for the pipe (<strong>|</strong>).
+            <dt>-c:v libx264</dt><dd>sets the codec as H.264</dd>
```
retokromer: I would say: video codec.

index.html
```diff
@@ -365,6 +365,45 @@
...
+            <dt>-c:a copy</dt><dd>audio remains as-is (no re-encode)</dd>
+            <dt><i>output_file.mp4</i></dt><dd>path and name of the output file.</dd>
```
retokromer: In the other recipes, I guess, the final period is only be used when a full sentence is closed.

retokromer (changes requested):
And thanks also for housekeeping.

kieranjol:
Apart from the issues that Reto found, this looks great to me, including the cleanup on other recipes.

kfrn:
Cool, will make those changes. Regarding `<br>` - happy to change, but I thought that was the proper HTML5 use? [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br) doesn't include a closing slash.

retokromer:
You are right, it's often used without, but it's an empty tag... (We could change all.)

retokromer:
[It depends if you wish to be able to parse with XML or not.]

kfrn:
I will stay consistent with the other entries :)
Now edited, hopefully I caught everything!

retokromer (approved):

retokromer:
Thank you very much!

kfrn:
Thank you both for reviewing!

ablwr:
Yes! Valid or not in HTML, it would throw an error if following the validation rules of XML, which requires a / as closing signifier either in a closing tag following an opening tag or a self-closing single tag.

Issue #115, opened 2016-12-03
Join Files Together - Clarifications needed


metacynicv2:
Heya,

Has anyone successfully concat files using this method? https://amiaopensource.github.io/ffmprovisr/#join_files

The language is a little unclear and I have not had success. 

Attached is how I interpreted how to create the join list, but I keep getting the also attached error. 

If there are any other possibly more simple methods for successfully joining DV files I'd be in love. 

![screen shot 2016-12-02 at 5 49 16 pm](https://cloud.githubusercontent.com/assets/9436241/20855835/b9695a7a-b8b7-11e6-9da4-048be249786a.png)

[BAVC1004511_2385M.txt](https://github.com/amiaopensource/ffmprovisr/files/628627/BAVC1004511_2385M.txt)

kfrn:
Hmm, it looks alright. I think I wrote or amended that command, so I know that it worked for me. Could using the curved quote marks (**‘’**) rather than straight apostrophes (**'**) be throwing things off?

metacynicv2:
Hi @kfrn - you are perfect. Changing the apostrophes from curvy to straight worked. Thank you!

kfrn:
Glad it worked! I'll add a note into the recipe too :)

retokromer:
Indeed, never use typographic style for commands! Some text editors do this by default, but it should be possible to disable this «feature».

PR #116, opened 2016-12-03
Add clarification to join files command

kfrn:
Added note clarifying punctuation to be used in the concat txt, after it came up [here](https://github.com/amiaopensource/ffmprovisr/issues/115).

### DIFF
--- a/index.html
+++ b/index.html
@@ -965,7 +965,7 @@
 file '<i>./second_file.ext</i>'
 . . .
 file '<i>./last_file.ext</i>'</pre>
-In the above, <strong>file</strong> is simply the word "file". Straight apostrophes ('like this') rather than curved quotation marks (‘like this’) should be used to enclose the file paths.<br/>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -965,7 +965,7 @@
 file '<i>./second_file.ext</i>'
 . . .
 file '<i>./last_file.ext</i>'</pre>
-In the above, <strong>file</strong> is simply the word "file".<br/>
+In the above, <strong>file</strong> is simply the word "file". Straight apostrophes ('like this') rather than curved quotation marks (‘like this’) should be used to enclose the file paths.<br/>
```
retokromer: I would say `must` instead of `should`, because this is compulsory – but I am not an English speaker.

retokromer:
Thank you, @kfrn!

index.html
kfrn: I would read them much the same - but better to be 100% clear. Thank you @retokromer :)

retokromer: The standardisation bodies make a distinction between MUST and SHOULD, therefore my doubt.

kfrn: 'Must' is more imperative, certainly. I defer to your knowledge of standards. ;)

todrobbins:
👍🏻👍🏻

PR #117, opened 2016-12-03
HTML 5.1

retokromer:
As we are using the «lazy» tag version only for `<br/>` but not for `<meta ...>` and `<link ...>`, I suggest to use only the official HTML 5.1 writing. I noticed that modern XML parsers can handle also actual HTML. (This answers a question asked, I guess, by @kfrn some time ago.)


todrobbins (approved):
Makes sense to me. 👍🏻👍🏻

retokromer:
Thank you, @todrobbins!

Issue #72, opened 2016-07-09
Interlaced ProRes


retokromer:
At a some point, I guess, we should also consider an example of interlaced ProRes.

kieranjol:
Good idea, especially considering the benefits of Kostya's encoder in
regards to interlacement
https://github.com/mediamicroservices/mm/commit/c022a1104e8f3b384273fbde1de8be2293a5c405

retokromer:
If I remember correctly, the `-c:v prores_ks` is only good for interlaced, but for progressive it's slower as `-c:v prores`. Therefore we should use `prores` "as default" and `prores_ks` only for interlaced. I also guess that ProRes produced today should be progressive, but for legacy we need indeed to deal with interlaced. Please correct me if I am wrong.

retokromer:
Thank you, @kieranjol, this is what https://github.com/mediamicroservices/mm/commit/c022a1104e8f3b384273fbde1de8be2293a5c405 (and the link there) says. Yet I will do some benchmark, as I am still feeling we should change only for interlaced, and not as a general rule for ProRes, because of the performance.

retokromer:
Hmm... at the 1st benchmark the `prores_ks` says:

```
[prores_ks @ 0x690180] Packet too small: is 237874, needs 275122 (slice: 3396). Correct allocation is not implemented. Update your FFmpeg version to the newest one from Git. If the problem still occurs, it means that your file has a feature which has not been implemented.
[prores_ks @ 0x690180] If you want to help, upload a sample of this file to ftp://upload.ffmpeg.org/incoming/ and contact the ffmpeg-devel mailing list. (ffmpeg-devel@ffmpeg.org)
```

kieranjol:
Hi Reto, I totally agree on only using `prores_ks` for interlaced items. I must admit, for access purposes, I've used the regular prores encoder and I haven't found any issues for TFF files, or clients haven't mentioned any either. I haven't done much rescaling though.

retokromer:
Does anyone know how (un)stable `prores_ks` is today? Both my interlaced and my 4:4:4 files seem not to be (completely) correct… 👎

kieranjol:
That is worrying. @dericed is using it at work I think, so perhaps he can chime in? What's the issue with the files?

retokromer:
I will test with 4K resonance patterns (or parts of it).

retokromer:
[...]

PR #121, opened 2016-12-31
update Transcode to ProRes

retokromer:
This should resolve (at least part of) https://github.com/amiaopensource/ffmprovisr/issues/72.

### DIFF
--- a/index.html
+++ b/index.html
@@ -93,7 +93,12 @@
             <dt><i>output_file</i></dt><dd>path, name and extension of the output file<br>
             The extension for the QuickTime container is <code>.mov</code>.</dd>
           </dl>
-          <p class="link"></p>
+          <p>FFmpeg comes with more than one ProRes encoder:</p>
+          <ul>
+            <li><code>prores</code> is much faster, can be used for progressive video only, and seems to be better for video according to Rec. 601 (Recommendation ITU-R BT.601).</li>
+            <li><code>prores_ks</code> generates a better file, can also be used for interlaced video, allows also encoding of ProRes 4444 (<code>-c:v prores_ks -profile:v 4</code>), and seems to be better for video according to Rec. 709 (Recommendation ITU-R BT.709).</li>
+          </ul>
+         <p class="link"></p>
         </div>
       </div>
     </div>


todrobbins:
Looks good! Thanks @retokromer!

richardpl:
The prores_aw, current "prores" encoder is going to be removed.d

kieranjol:
Is it going to be removed completely, or will kostya's encoder be made the
default instead? That's what I picked up from your patch on ffmpeg-devel.

retokromer:
Thank you, @richardpl! We will certainly update our information as soon as a new release of FFmpeg that includes the change will be published.

For the moment, one problem I see is the big difference regarding the executing time.

richardpl:
Well that only applies for default parameters. Try with -qscale:v 4 for prores_ks.
With default parameters prores_ks try very hard to match bitrate.

retokromer:
@richardpl: I cannot find a transparent documentation about the exact effect of the `-qscale` parameter, but I notice «strange» behaviours when applied to the resonance test patterns I usually use to check the encoding quality. Do you have a reference where this is documented? Otherwise I will delve into the code and do the reverse-engineering, as soon as I will have the time to do so.

(I almost never use the default parameters, but I do measure a considerable difference in executing time between `prores` and `prores_ks`.)

richardpl:
It is documented in source code. Let me cite original author:
"The answer is that the encoder tries to find the best quantiser for each slice where best = gives the least error and uses about the allocated number of bits. I used simple linear search and IIRC didn't implement good handling for the (relatively) low bitrates because it worked GoodEnough(TM). Though some people said it actually followed bitrate "better" (i.e. with less variation)."

retokromer:
Thank you, @richardpl!

I actually would like to know more regarding what happens exactly when switching e.g. from 4 to 5. I also noticed that in the different web documents, there are some «fluctuations» 0–10 (10 best), 1–32 (1 best) or 0–32 (0 best). My guess is that the interval depends on the bit-depth (BTW: we would also need 12, therefore yuva444p10le seem not a good choice to me, and 16 for alpha). I will check the code.

richardpl:
16 for alpha is by default. 12bit is snake oil until someone provides actual 12 bit sample.

retokromer:
Oh, I can provide 4444 with 12 bit as soon as I am back in the lab.

retokromer:
@richardpl Please give me an [email address](https://reto.ch/cgi-bin/feedback.pl) where I can sent you the link to the files.

PR #125, opened 2017-01-26
delete `-filter:v` for `ffplay`

retokromer:
The syntax of `ffplay` differs from the syntax of `ffmpeg`. In our case, `-filter:v` does not exist in `ffplay` at all. (Don’t ask why.)


ablwr:
Ahh! Thank you!

PR #127, opened 2017-01-27
adds qctool report scripts

ablwr:
CC @kieranjol (whom I can't tag for some reason) for Windows compatibility concerns.

This doesn't explain the huge bulk of the script, but maybe one day?

Thanks to @dericed the patron saint of everything!

### DIFF
--- a/index.html
+++ b/index.html
@@ -964,6 +964,66 @@
   </div>
   <!-- ends Check interlacement   -->
 
+  <!-- QCTools Report -->
+  <span data-toggle="modal" data-target="#qctools"><button type="button" class="btn btn-default" data-toggle="tooltip" data-placement="bottom" title="This will create q QCTools report for a video file with audio track">QCTools report (with audio)</button></span>
+  <div id="qctools" class="modal fade" tabindex="-1" role="dialog">
+    <div class="modal-dialog modal-lg">
+      <div class="modal-content">
+        <div class="well">
+          <h3>Creates a QCTools report</h3>
+          <p><code>ffprobe -f lavfi -i "movie=<i>input_file</i>:s=v+a[in0][in1],[in0]signalstats=stat=tout+vrep+brng,cropdetect=reset=1:round=1,idet=half_life=1,split[a][b];[a]field=top[a1];[b]field=bottom,split[b1][b2];[a1][b1]psnr[c1];[c1][b2]ssim[out0];[in1]ebur128=metadata=1,astats=metadata=1:reset=1:length=0.4[out1]" -show_frames -show_versions -of xml=x=1:q=1 -noprivate | gzip > <i>input_file</i>.qctools.xml.gz</code></p>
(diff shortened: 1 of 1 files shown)


kieranjol:
Ah great! Yes, an explanation of the signnalstats command could come later. Would have to look at other examples of how to break it up gracefully.

index.html
```diff
@@ -964,6 +964,66 @@
...
+            <dd>This very large lump of commands declares the input file and passes in a request for all potential data signal information for a file with one video and one audio track</dd>
+            <dt>-show_frames</dt><dd>asks for information about each frame and subtitle contained in the input multimedia stream</dd>
+            <dt>-show_versions</dt><dd>asks for information related to program and library versions</dd>
+            <dt>-of xml=x=1:q=1</dt><dd>sets the data export format to XML </dd>
+            <dt>-noprivate</dt><dd>hides any private data that might exist in the file</dd>
+            <dt>| gzip</dt><dd>The | is to "pipe" (or push) the data into a compressed file format</dd>
+            <dt><code>></code></dt><dd>redirects the standard output (the data made by ffprobe about the video) </dd>
+            <dt><i>input_file</i>.qctools.xml.gz</dt><dd>names the zipped data output file, which can be named anything but needs the extension qctools.xml.gz for compatibility issues</dd>
```
kieranjol, line 988: Should this be output file?

index.html
```diff
@@ -964,6 +964,66 @@
...
```
kieranjol, line 1018: Again, output_file? Would have to changed in the full command further up too. Aside from that, all looks great !

index.html
retokromer: No, I guess it’s meant to have the same name as the input file, but with `.qctools.xml.gz` appended which makes sense to me.

retokromer: No, I guess it’s meant to have the same name as the input file, but with `.qctools.xml.gz`  appended which makes sense to me.

retokromer:
Thank you, LGTM

index.html
ablwr: You are right, it *could* be output file but I did as @retokromer said, wanted to use the same as the input file (just as good practice).

ablwr:
I'll make an issue for expanding on the nuances in the signalstats zone but merge for now so people can see the script (I think most people don't realize this can be done at all)

kieranjol (commented):
Ah yes I get you. You're creating a sidecar, so sorry about that.

dericed:
Could you add a cross-reference to the corresponding documentation in the qctools repo: https://github.com/bavc/qctools/blob/master/docs/data_format.md#creating-a-qctools-document.

ablwr:
@dericed do you prefer that URL or http://bavc.github.io/qctools/data_format.html ?

retokromer:
I added @dericed’s but can change, of course: https://github.com/amiaopensource/ffmprovisr/pull/133

ablwr:
thank you Reto!

PR #133, opened 2017-01-27
add cross ref to QCTools doc

retokromer:


ablwr:
Best at the bottom or up top before the explanation?

retokromer:
I don’t think before the explanation, but perhaps just after it, instead as of a «footnote».

ablwr:
Before the breakdown of each part of the command, I mean? This breakdown will get very lengthy when the signalstats part gets parsed.

ablwr:
I think it'd be best as a new <p> immediately under "This will create an XML report for use in QCTools for a video file with one video track and one audio track."

But also don't care that much so do as you think is best!

retokromer:
Moved.

PR #136, opened 2017-01-28
Embed links; remove extra line about comma

kfrn:
A small tidy-up.
I haven't got the OCR ones working yet, but these scripts are mega cool! Thanks for adding them.


ablwr:
Oh, the OCR ones as I found out on twitter from @privatezero need to be installed `--with-tesseract`! I forgot to make a note to add that.

kfrn:
Ah, thanks! I think it would be great if you or @privatezero added that in. :)

ablwr:
I'll add it!

retokromer:
Thank you, @kfrn! It’s indeed more readable.

retokromer:
We use it for intertitles – also when written in [Fraktur](https://en.wikipedia.org/wiki/Fraktur) – and subtitle extraction. It works well!

PR #142, opened 2017-02-02
adds link to gif for audio graph

ablwr:
@privatezero added this gif but it was never added to the script it relates to. This adds it.


retokromer:
LGTM

PR #143, opened 2017-02-06
correct astats example

privatezero:
the example for astats was actually a gif of showspectrum (an as yet to be added script).  Added and edited to correct example!


ablwr:
looooooooool

Issue #141, opened 2017-01-31
deinterlacing command (+question)


bturkus:
the @privatezero tutorial last week got me thinking about deinterlacing. the only yadif flag in ffmprovisr is in the "Transcode to Prores" command.

Wondering if there should be something separate for clarity, and wondering if anyone has any exp with deinterlacing filters other than yadif (I count 4 others: bwdif, kerndeint, nnedi, w3fdif).

Alternatively, does anyone have a method for controlled deinterlacing (or denoising) testing? 

thanks.

kieranjol:
+1 for separating the yadif command from the prores recipe (if that's what
you meant)

ablwr:
you are great / no idea

privatezero:
Hey! Not sure if this is what you were thinking, but when I was looking into detelecine filters I used qctools to compare/analyze results. I would set it to show me the full image on one side and the field split on the other and then advance frame by frame to see what effect the filter was having. Seems like this approach could work to compare the effects of deinterlace filters!

retokromer:
I did some quick tests in the past, I guess five years ago or so. And this should indeed be analysed again and in-depth!

At that time, the default bilinear was clearly the fastest and yadif having the best ratio between speed and quality. We have used internally a self-made filter for one single project, because the artist was unhappy with all the existing filters. It gave an excellent quality, but it was much slower than yadif. We currently use almost always yadif.

A little similar question also rises about the variants of 4:2:0 subsampling.

kfrn:
Related, I had been thinking about an IVTC filter. Many use AVISynth for IVTC, but I'm experimenting with ffmpeg for this.  
Although, this is more for DVD-ripping than a likely archival use case.

retokromer:
This is – at least partially – resolved by https://github.com/amiaopensource/ffmprovisr/pull/151.

retokromer:
Some additional work on https://github.com/amiaopensource/ffmprovisr/pull/151 will be done during the next days, but I guess we are almost done. Therefore I close this one.

PR #151, opened 2017-02-26
Add deinterlace recipe

kfrn:
Before
interlaced_video

and after
deint_video

### DIFF
--- a/index.html
+++ b/index.html
@@ -445,6 +445,38 @@
   </div>
   <!-- ends Transcode to H.265 -->
 
+  <!-- Deinterlace video -->
+  <span data-toggle="modal" data-target="#deinterlace"><button type="button" class="btn btn-default" data-toggle="tooltip" data-placement="bottom" title="Deinterlace video">Deinterlace video</button></span>
+  <div id="deinterlace" class="modal fade" tabindex="-1" role="dialog">
+    <div class="modal-dialog modal-lg">
+      <div class="modal-content">
+        <div class="well">
+          <h3>Deinterlace a video</h3>
+          <p> <code>ffmpeg -i <i>input_file</i> -c:v libx264 -vf "yadif,format=pix_fmts=yuv420p" <i>output_file</i></code> </p>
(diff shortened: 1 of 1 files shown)


ablwr:
WHHHOOOAAAAAA

retokromer:
Thank you very much indeed!

ablwr:
I'd be interested in seeing those two images integrated into the display of the script, like a couple of the filters now have.

retokromer:
This was the 100<sup>th</sup> pull request!

ablwr:
We must celebrate!

kieranjol:
Brilliant!! And those images are great.

kfrn:
Shall I embed the images, then?

ablwr:
I would like this!

ghost:
`format=pix_fmts=yuv420p` can be shortened to `format=yuv420p`, and `yadif=1` often looks better than the default mode `0` depending on your content. Note that mode `1` will double the output frame rate which is expected because it will output one frame for each field. Also, there are other deinterlacers available to try such as [bwdif](https://ffmpeg.org/ffmpeg-filters.html#bwdif) and [w3fdif](https://ffmpeg.org/ffmpeg-filters.html#w3fdif).

kfrn:
Thank you @llogan. I went to the ffmpeg docs for that ([here](https://www.ffmpeg.org/ffmpeg-all.html#Examples-89)), so thanks for letting me know about the shorter version.
Perhaps I should add a note about the different options for `yadif`.
There was some talk in #141 about comparing different deinterlacers. Here I just went with `yadif` since it seems widely used, and I liked the results.

retokromer:
Yes, @kfrn, you could possibly add the note. And then I’ll make a new release.

I will also try to have my students testing the current FFmpeg’s implementation of the de-interlacing filters next week, when I‘ll be teaching at the Bern University of Applied Sciences. That’s an interesting exercise for them, which could provide useful information to `ffmprovisr` as well!

kfrn:
Is it okay if I do this in the next few days? I also wanted to add one or two new recipes, but they're not finished yet.
And very cool, @retokromer ! Please do report back. :)

PR #157, opened 2017-02-28
Fix links

kfrn:
I sneaked in a couple of other link changes before, @retokromer


retokromer:
Two remarks:

- Is the `index.html` part not requested?
- Can we avoid the full link as you did previously? (This way it works off-line as well.)

retokromer:
I don’t know if `index.html` or `./index.html` is better.
@kieranjol : Is there any Windows issue?

kfrn:
Ah bugger. I just grabbed them from the browser address. Now fixed - sorry!

retokromer:
Thank you! I guess we can merge.

retokromer:
@kfrn: It works fine now on different browsers, except one. Therefore it’s actually the browser and not the code. Thank you again!

kfrn:
I'm on my Windows computer. Working offline, either seem to work - but only when opening in a new tab. Neither go to the linked recipe just from clicking.

retokromer:
Thank you for the feedback! I just made a release, and will test it on my Intel NUC test machine under Windows (for the training in Bangkok).

PR #159, opened 2017-03-02
adds audio testfile (sine wave)

ablwr:
Shouldn't just be video test files up in here!

### DIFF
--- a/index.html
+++ b/index.html
@@ -1354,7 +1354,7 @@ foreach ($file in $inputfiles) {
       <div class="modal-content">
         <div class="well">
           <h3>Sine wave</h3>
-          <p>Generate a test audio file playing a Sine wave.</p>
+          <p>Generate a test audio file playing a sine wave.</p>
           <p><code>ffmpeg -f lavfi -i "sine=frequency=1000:sample_rate=48000:duration=5" -c:a pcm_s16le <i>output_file</i>.wav</code></p>
           <dl>
             <dt>ffmpeg</dt><dd>starts the command</dd>
--- a/index.html
+++ b/index.html
@@ -1233,7 +1233,7 @@ foreach ($file in $inputfiles) {
 
   </div>
   <div class="well">
-  <h4>Test videos</h4>
+  <h4>Test files</h4>
 
   <!-- Mandelbrot -->
   <span data-toggle="modal" data-target="#mandelbrot"><button type="button" class="btn btn-default" data-toggle="tooltip" data-placement="bottom" title="Make a mandelbrot test pattern video">Mandelbrot</button></span>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1347,6 +1347,29 @@
...
+          <p>Generate a test audio file playing a Sine wave.</p>
+          <p><code>ffmpeg -f lavfi -i "sine=frequency=1000:sample_rate=48000:duration=5" -c:a pcm_s16le <i>output_file</i>.wav</code></p>
+          <dl>
+            <dt>ffmpeg</dt><dd>starts the command</dd>
+            <dt>-f lavfi</dt><dd>tells ffmpeg to use the libavfilter input virtual device <a href="http://ffmpeg.org/ffmpeg-devices.html#lavfi" target="_blank">[more]</a></dd>
+            <dt>-i "sine=frequency=1000:sample_rate=48000:duration=5"</dt><dd>Sets the signal to 1000 Hz, sampling at 48kHz, and for 5 seconds</dd>
+            <dt>-c:a pcm_s16le</dt><dd>encodes the audio codec in <code>pcm_s16le</code> (the default encoding for wav files). pcm represents pulse-code moderation format (raw bytes), <code>16</code> means 16 bits per sample, and <code>le</code> means "little endian"</dd>
+            <dt><i>output_file</i>.wav</dt><dd>path, name and extension of the output file</dd>
+          </dl>
+          <p class="link"></p>
```
retokromer: You could mention in a note, that there are many possibilities in `lavfi`, e.g. you can evaluate any mathematical function:
```
ffmpeg -f lavfi -i aevalsrc="sin(440*2*PI*t)" -t 10 la.wav
```

retokromer: Sorry, I guess the note «la» is called «A» in English.

index.html
```diff
@@ -1347,6 +1347,29 @@
...
+      <div class="modal-content">
+        <div class="well">
+          <h3>Sine wave</h3>
+          <p>Generate a test audio file playing a Sine wave.</p>
+          <p><code>ffmpeg -f lavfi -i "sine=frequency=1000:sample_rate=48000:duration=5" -c:a pcm_s16le <i>output_file</i>.wav</code></p>
+          <dl>
+            <dt>ffmpeg</dt><dd>starts the command</dd>
+            <dt>-f lavfi</dt><dd>tells ffmpeg to use the libavfilter input virtual device <a href="http://ffmpeg.org/ffmpeg-devices.html#lavfi" target="_blank">[more]</a></dd>
+            <dt>-i "sine=frequency=1000:sample_rate=48000:duration=5"</dt><dd>Sets the signal to 1000 Hz, sampling at 48kHz, and for 5 seconds</dd>
+            <dt>-c:a pcm_s16le</dt><dd>encodes the audio codec in <code>pcm_s16le</code> (the default encoding for wav files). pcm represents pulse-code moderation format (raw bytes), <code>16</code> means 16 bits per sample, and <code>le</code> means "little endian"</dd>
```
privatezero: Just a nit-pick! There is a slight typo- 'Modulation' not 'moderation'

privatezero: Word. Audiooooooo

retokromer: More nit-picking: I would unify: either 1000 Hz and 48 kHz (my favourite) or 1000Hz and 48kHz.

ablwr: Whooooops late-night typing. Correcting both.

index.html
ablwr: Ahh I was trying to keep with the exact pattern we use for the other recipes? Maybe an example of this merits its own recipe!

index.html
```diff
@@ -1347,6 +1347,29 @@
...
   <!-- ends Play VGA SMPTE bars -->
 
+  <!-- Sine wave -->
+  <span data-toggle="modal" data-target="#sine_wave"><button type="button" class="btn btn-default" data-toggle="tooltip" data-placement="bottom" title="Generate a test audio file playing a sine wave">Sine wave</button></span>
+  <div id="sine_wave" class="modal fade" tabindex="-1" role="dialog">
+    <div class="modal-dialog modal-lg">
+      <div class="modal-content">
+        <div class="well">
+          <h3>Sine wave</h3>
+          <p>Generate a test audio file playing a Sine wave.</p>
```
retokromer: Does sine wear uppercase in English?

index.html
retokromer: If another recipe, then we could do something more crazy and combine sounds together.

retokromer: We could include a recipe around `aevalsrc`.

ablwr:
Nooooo idea, hence my wavering! I think not.

retokromer:
According to my books, without capital.

retokromer:
:1st_place_medal:

PR #162, opened 2017-03-06
typography

retokromer:
- modify `'` --> `’` in text
- modify `" … "` --> `“ … ”` in text
- modify `http` --> `https` when possible
- unify newline after `<br>`
- add note `-vf is an alias of -filter:v` in the ffmpeg commands


ablwr:
I do prefer dumb quotes, even in text, because it's difficult to naturally create them using a text editor intended for code.

retokromer:
Hmm… It’s just a parameter in the editor. Yet we can unify all to dump, if preferred.

retokromer:
I cannot resolve the conflict without reintroducing an error. If the pull request is accepted, I will correct again.

@ablwr Can we decide if for `ffmprovisr` we prefer typographic style (which in my opinion fits the sophisticated layout) or dump apostrophes and quotes.

PR #166, opened 2017-03-31
Add more detail to H.264 and GIF commands

kfrn:
I hope I'm not stepping on any toes, but I thought it would be useful to add a little more detail. With the gif command (contributed by @retokromer, I think? Thank you!), it was to understand it better for myself more than anything - my text could be improved I think, so feel free to edit.

### DIFF
--- a/index.html
+++ b/index.html
@@ -153,9 +153,12 @@
           <p>In order to use the same basic command to make a higher quality file, you can add some of these presets:</p>
           <p><code>ffmpeg -i <i>input_file</i> -c:v libx264 -pix_fmt yuv420p -preset veryslow -crf 18 -c:a copy <i>output_file</i></code></p>
           <dl>
-            <dt>-preset <i>veryslow</i></dt><dd>This option tells ffmpeg to use the slowest preset possible for the best compression quality.</dd>
-            <dt>-crf <i>18</i></dt><dd>Specifying a lower CRF will make a larger file with better visual quality. 18 is often considered a “visually lossless” compression.</dd>
+            <dt>-preset <i>veryslow</i></dt><dd>This option tells ffmpeg to use the slowest preset possible for the best compression quality.<br>
+            Available presets, from slowest to fastest, are: <code>veryslow</code>, <code>slower</code>, <code>slow</code>, <code>medium</code>, <code>fast</code>, <code>faster</code>, <code>veryfast</code>, <code>superfast</code>, <code>ultrafast</code>.</dd>
+            <dt>-crf <i>18</i></dt><dd>Specifying a lower CRF will make a larger file with better visual quality. The scale ranges between 0-51, with 0 being lossless and 51 the worst possible quality.<br>
+              If no crf is specified, <code>libx264</code> will use a default value of 23. 18 is often considered a “visually lossless” compression.</dd>
           </dl>
+          <p>For more information, see the <a href="https://trac.ffmpeg.org/wiki/Encode/H.264" target="_blank">FFmpeg and H.264 Encoding Guide</a> on the ffmpeg wiki.</p>
           <p class="link"></p>
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -153,9 +153,12 @@
...
           <p><code>ffmpeg -i <i>input_file</i> -c:v libx264 -pix_fmt yuv420p -preset veryslow -crf 18 -c:a copy <i>output_file</i></code></p>
           <dl>
-            <dt>-preset <i>veryslow</i></dt><dd>This option tells ffmpeg to use the slowest preset possible for the best compression quality.</dd>
-            <dt>-crf <i>18</i></dt><dd>Specifying a lower CRF will make a larger file with better visual quality. 18 is often considered a “visually lossless” compression.</dd>
+            <dt>-preset <i>veryslow</i></dt><dd>This option tells ffmpeg to use the slowest preset possible for the best compression quality.<br>
+            Available presets, from slowest to fastest, are: <code>veryslow</code>, <code>slower</code>, <code>slow</code>, <code>medium</code>, <code>fast</code>, <code>faster</code>, <code>veryfast</code>, <code>superfast</code>, <code>ultrafast</code>.</dd>
+            <dt>-crf <i>18</i></dt><dd>Specifying a lower CRF will make a larger file with better visual quality. The scale ranges between 0-51, with 0 being lossless and 51 the worst possible quality.<br>
```
retokromer: [As long as I know](https://avpres.net/FFmpeg/im_H264.html) the interval [0..51] is valid only for 8-bit, but for 10-bit it’s [0..63].

index.html
```diff
@@ -819,7 +819,7 @@
...
             <dt>-filter_complex "fps=<i>frame rate</i>,scale=<i>width</i>:<i>height</i>,palettegen"</dt><dd>a complex filtergraph.<br>
               Firstly, the fps filter sets the frame rate.<br>
-              Then the scale filter resizes the image. You can specify both the width and the height, or specify a value for one and use a scale value of <i>-1</i> for the other to preserve the aspect ratio. (For example, <code>500:-1</code> would create a GIF 500 pixels wide and with a height proportional to the original video). In the first script above, <code>:flags:lanczos</code> specifies that the Lanczos rescaling algorithm will be used to resize the image.<br>
+              Then the scale filter resizes the image. You can specify both the width and the height, or specify a value for one and use a scale value of <i>-1</i> for the other to preserve the aspect ratio. (For example, <code>500:-1</code> would create a gif 500 pixels wide and with a height proportional to the original video). In the first script above, <code>:flags=lanczos</code> specifies that the Lanczos rescaling algorithm will be used to resize the image.<br>
```
retokromer: Did we not use capitals for the abbreviation, like TIFF, DCP, GIF, etc.? (No, GIF it’s not me :-)

index.html
kfrn: Ah, I had just looked to the [wiki](https://trac.ffmpeg.org/wiki/Encode/H.264). Thanks for the link. Perhaps I should add a note explaining that the 0-51 scales applies for `yuv420p` only.

index.html
kfrn: Will change!
The real question ... do you pronounce it with a hard or soft 'g'? ;)

retokromer (approved):

index.html
retokromer: With a hard, but perhaps this is not English…

retokromer:
Thank you!

index.html
kfrn: It's a matter of [great debate](https://en.wikipedia.org/wiki/GIF#Pronunciation_of_GIF), you see! I have always said it with a hard 'g'.

kfrn:
As always, thank you for your careful revisions Reto!

ablwr:
Haha I love when these things happen while I snooze!! Thank you so much @kfrn! I had felt the same about the GIF one... and also considered breaking it up officially into two commands. Thanks for these changes!

Issue #169, opened 2017-04-13
Suggestion: Splitting "change formats" into "change codec/container" and "change formats"?


pjotrek-b:
I caught myself wishing to separate the existing entries in "Change formats" into 2 categories:
  1. Change codec/container
  2. Change formats

I think that might make it easier and clearer for non-experienced users to find what they're looking for?
Just a suggestion.

If you agree, then I'd provide the patch.

retokromer:
I personally prefer to distinguish in a more pedantic way between

- change container
- change codec
- change format

But if it’s just me, then of course I’ll agree.

pjotrek-b:
Yes, absolutely agreed!
I loooove sorting things properly :grin:

pjotrek-b:
Hm... Ran into the first issue:
"_WAV to MP3_" is actually of both: "change container + codec"

But I'd suggest if transcoding happens, to put it under "change codec".
I'd also suggest to put only "-c copy" entries in "change container".

retokromer:
Agreed.

ablwr:
Also agree. Thanks @pjotrek-b, 👑 👑 👑

retokromer:
resolved in https://github.com/amiaopensource/ffmprovisr/pull/172 by @pjotrek-b (thank you :-)

PR #172, opened 2017-04-13
Split "change formats" into separate categories

pjotrek-b:
Related to suggestion #169.

The "mkv to mp4" rewrap command feels a bit alone now, but probably not for long :wink:

### DIFF
--- a/index.html
+++ b/index.html
@@ -417,6 +417,31 @@
   </div>
   <!-- ends 4:3 to 16:9 -->
 
+  <!-- 16:9 to 4:3 -->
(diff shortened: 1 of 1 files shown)


retokromer (approved):

retokromer:
It’s fine that way!

index.html
```diff
@@ -476,6 +311,188 @@
   </div>
   <!-- ends Transcode to H.265 -->
 
+  <p>&nbsp;</p>
```
retokromer, line 314: Why is this necessary?

retokromer (commented):
@pjotrek-b I missed this:

index.html
pjotrek-b: To make an optical gap between video and audio-only commands.
First wanted to introduce an "audio" headline, but that looked bloated...

PR #173, opened 2017-04-15
add script for comparing audio framemd5

privatezero:
Was linking to the script for comparing video framemd5s, made some tweaks so that it will work for audio comparison.

Also, is it worth explicitly noting in the explanation that ffmpeg by default converts the input to 16 bit when it generates the framemd5s, so the hashes are for that value and not the actual content of the input file?


ablwr:
LGTM but maybe someone with more area expertise should review.

check_audio_framemd5.sh
```diff
@@ -0,0 +1,77 @@
...
+[[ -z "${#}" || ! ${input_file} || ! ${input_hash} ]] && _output_prompt
+echo -e "${BLUE}Please wait...${NC}"
+unset md5_tmp
+if [[ $OSTYPE = "cygwin" ]]; then
+    md5_tmp=""${USERPROFILE}/$(basename ${input_hash}).tmp""
+else
+    md5_tmp="${HOME}/$(basename ${input_hash}).tmp"
+fi
+# Find audio frame size for hash calculation
+sample_rate=$(grep -v '^#' ${input_hash} | head -n 1 | tr -d ' ' | cut -d',' -f4)
```
privatezero: the main change is in setting and unsetting this variable, and then to the ffmpeg command

check_audio_framemd5.sh
```diff
@@ -0,0 +1,77 @@
...
+        :) echo -e "${RED}Error: option -${OPTARG} requires an argument${NC}" ; _output_prompt ;;
+        *) echo -e "${RED}Error: bad option -${OPTARG}${NC}" ; _output_prompt ;;
+    esac
+done
+
+[[ -z "${#}" || ! ${input_file} || ! ${input_hash} ]] && _output_prompt
+echo -e "${BLUE}Please wait...${NC}"
+unset md5_tmp
+if [[ $OSTYPE = "cygwin" ]]; then
+    md5_tmp=""${USERPROFILE}/$(basename ${input_hash}).tmp""
```
dericed: quadruple quotes?!

check_audio_framemd5.sh
dericed: should put $input_hash in quotes throughout, else when you run `check_audio_framemd5.sh -m " ; find ~/Music -name *.mp3 -delete"` you might be sad.

check_audio_framemd5.sh
privatezero: No idea! Could it have something to do with how Cygwin works?

check_audio_framemd5.sh
privatezero: hahahaha ooh yeah good point! I'll take a look at this and the other script that is on ffmprovisr and make that change.

check_audio_framemd5.sh
```diff
@@ -0,0 +1,77 @@
...
+  ${SCRIPT} -h
+    This help.
+  ${SCRIPT} -i <av_file> -m <md5_file>
+    Pass to the script the audio-visual file and the corresponding MD5
+    file to check.
+Dependency:
+  ffmpeg
+About:
+  Version: ${VERSION}
+  Website: https://github.com/amiaopensource/ffmprovisr/blob/gh-pages/check_framemd5.sh
```
retokromer: URL should read check_audio_framemd5.sh

check_audio_framemd5.sh
retokromer: It’s double quote for Cygwin, not quadruple.

index.html
```diff
@@ -1201,7 +1201,7 @@
             <dt>-vn</dt><dd>ignores the video stream (video no)</dd>
             <dt><i>output_file</i></dt><dd>path, name and extension of the output file</dd>
           </dl>
-          <p>You may verify an MD5 checksum file created this way by using a <a href="check_framemd5.sh" target="_blank">Bash script</a>.</p>
+          <p>You may verify an MD5 checksum file created this way by using a <a href="check_audio_framemd5.sh" target="_blank">Bash script</a>.</p>
```
retokromer, line 1205: Good catch!

check_audio_framemd5.sh
```diff
@@ -0,0 +1,77 @@
...
+echo -e "${BLUE}Please wait...${NC}"
+unset md5_tmp
+if [[ $OSTYPE = "cygwin" ]]; then
+    md5_tmp=""${USERPROFILE}/$(basename ${input_hash}).tmp""
+else
+    md5_tmp="${HOME}/$(basename ${input_hash}).tmp"
+fi
+# Find audio frame size for hash calculation
+sample_rate=$(grep -v '^#' ${input_hash} | head -n 1 | tr -d ' ' | cut -d',' -f4)
+$(ffmpeg -i ${input_file} -loglevel 0 -filter_complex "asetnsamples=n='$sample_rate'" -f framemd5 -vn ${md5_tmp})
```
retokromer: `-filter:a` should fit, instead of `-filter_complex`.

check_audio_framemd5.sh
retokromer: If I remember carefully, the other script don’t need the extraction of metadata.

I suggest that we change the name of the existing script to `check_video_framemd5.sh` that checks video only, and we add a new script `check_framemd5.sh` that checks both video and audio (as well as a new recipe that generates both).

retokromer (changes requested):

retokromer (changes requested):
Please update the date of `check_video_framemd5.sh`:
`VERSION='2017-04-16'`

retokromer (approved):

retokromer (approved):

retokromer:
Thank you!

PR #176, opened 2017-04-17
Add examples for using new Signature filter

privatezero:
Examples for using Signature filter for file comparison and for sidecar generation.


retokromer (changes requested):
We use `FFmpeg` for the full project and `ffmpeg` for one of its function. Could you possibly change to small caps?

retokromer (approved):

retokromer:
Thank you!

Issue #150, opened 2017-02-26
[wishes] 1) Brew formula, 2) CLI cheat-sheet and 3) GUI for ffmprovisr


retokromer:
This way, any time
```
brew update
brew upgrade
```
is run, also `ffmprovisr` will be updated to the lastest release, if needed.

ablwr:
What do you have in mind for this, just that it clones and pulls down the most recent version and someone can open up index.html locally by running `ffmprovisr`?

One of my colleagues asked me if I'd considered converting to a CLI cheat-sheet interface, which I think is interesting but unsure how we would be able to realistically do that, although if that's something you're interested in, I'm happy to brainstorm in this thread about it.

kieranjol:
What was their CLI idea? Would it function as a kind of frontend for ffmpeg
or would it just give you info?

ablwr:
Something like a Ruby gem that would act as a simplified `man` page, so you could call it and it would pull up the names of recipes, and then you can, like, pass in 4 for this recipe and it would give you the line. He was thinking about something like this: http://cheat.errtheblog.com/ for short, concise documentation for when you need a memory-jog without having to leave the Terminal. This is my #1 usecase for ffmprovisr.

retokromer:
I was thinking just in downloading the folder and you can open `index.html` locally with the browser. You don’t need internet access for using it. Yet it’s automatically updated every time the Homebrew stuff is updated. That’s my main point, related to my activity in countries with sometimes catastrophic access to the Web.

ablwr:
Yes, making sure ffmprovisr opened and worked fine locally was a big consideration when I started this project!

retokromer:
The update of the local copy would be automated (e.g. with `vrecord` or any other Homebrew formula).

retokromer:
Yet the idea of having a CLI cheat-sheet is more than interesting.

ablwr:
I've also been experimenting with Electron.js to turn ffmprovisor into a functioning GUI, although I'm still many many hours away from a functional prototype.

ablwr:
So this is a thing, but it is a tedious thing to keep building...

<img width="757" alt="screen shot 2017-03-10 at 19 23 46" src="https://cloud.githubusercontent.com/assets/3260492/23818730/1e968060-05cc-11e7-85b5-8156244b6558.png">

retokromer:
I modified the title, because we are discussing here three distinct things:

[...]

PR #181, opened 2017-05-04
Script to open ffmprovisr when installed via homebrew

privatezero:
To partially address https://github.com/amiaopensource/ffmprovisr/issues/150 I made a homebrew formula at https://github.com/amiaopensource/homebrew-amiaos/blob/ffmprovisr/ffmprovisr.rb.

This simple script would provide an easy shortcut to open the local ffmprovisr install via terminal.  Only supports homebrew/macOS but would be easy to add in a path for linuxbrew.

If this is merged/new release made then the brew formula could be updated and merged (currently would error since this script doesn't exist in the most recent release).

* revised to remove apology as per https://github.com/amiaopensource/ffmprovisr/issues/182

### DIFF
--- a/scripts/ffmprovisr
+++ b/scripts/ffmprovisr
@@ -9,3 +9,13 @@ if [[ $OSTYPE = darwin* ]] ; then
     fi
     open "${ffmprovisr_path}"
 fi
+
+if [[ $OSTYPE = linux-gnu ]] ; then
+    if [ -d ~/.linuxbrew/Cellar/ffmprovisr ] ; then
+		ffmprovisr_path=$(find ~/.linuxbrew/Cellar/ffmprovisr -iname 'index.html' | sort -M | tail -n1)
+    fi
+    if [ -z "${ffmprovisr_path}" ] ; then
+       		ffmprovisr_path='https://amiaopensource.github.io/ffmprovisr/'
+    fi
+    xdg-open "${ffmprovisr_path}"
+fi
--- a/scripts/ffmprovisr
+++ b/scripts/ffmprovisr
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+if [[ $OSTYPE = darwin* ]] ; then
+    if [ -d /usr/local/Cellar/ffmprovisr ] ; then
+        ffmprovisr_path=$(find /usr/local/Cellar/ffmprovisr -iname 'index.html' | sort -M | tail -n1)
+    fi
+    if [ -z "${ffmprovisr_path}" ] ; then
+        ffmprovisr_path='https://amiaopensource.github.io/ffmprovisr/'
+    fi
+    open "${ffmprovisr_path}"
+fi
--- a/scripts/ffmprovisr
+++ b/scripts/ffmprovisr
@@ -8,8 +8,7 @@ if [[ $OSTYPE = darwin* ]] ; then
         ffmprovisr_path='https://amiaopensource.github.io/ffmprovisr/'
     fi
     open "${ffmprovisr_path}"
-elif
- [[ $OSTYPE = linux-gnu ]] ; then
+elif [[ $OSTYPE = linux-gnu ]] ; then
     if [ -d ~/.linuxbrew/Cellar/ffmprovisr ] ; then
         ffmprovisr_path=$(find ~/.linuxbrew/Cellar/ffmprovisr -iname 'index.html' | sort -M | tail -n1)
     fi
--- a/scripts/ffmprovisr
+++ b/scripts/ffmprovisr
@@ -12,10 +12,10 @@ fi
 
 if [[ $OSTYPE = linux-gnu ]] ; then
     if [ -d ~/.linuxbrew/Cellar/ffmprovisr ] ; then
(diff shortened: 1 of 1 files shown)


ablwr:
Never apologize for contributions!!! ⚔️

retokromer:
The path for Linux should be easy to implement with an `elif` into the test. I don’t remember it for sure and I cannot test it now, because I am still travelling in the USA… and have only macOS in my luggage.

privatezero:
yeah, I figured I could add linux support in with my home laptop when I get the time. Should just need a revised path and then a replacement for `open` which (correct me if I am wrong) is osx specific?

retokromer:
In my memory, just needs the alternative path, and `open` is correct in Linux.
(alternative as used in the pre-*** times)

privatezero:
At least in Ubuntu `open` threw an error (seems to be a command related to virtual terminals).  Added a conditional to use `xdg-open` and the linuxbrew Cellar path and things appear to be hunky-dory for Ubuntu!

kieranjol:
Off the top of my head, the command for ubuntu 14 is
gnome-open. I think it changed to
Xdg-open for ubuntu 16.

Cant confirm cos my laptop is on a ✈


On 4 May 2017 4:16 p.m., "Andrew Weaver" <notifications@github.com> wrote:

At least in Ubuntu open threw an error (seems to be a command related to
virtual terminals). Added a conditional to use xdg-open and the linuxbrew
Cellar path and things appear to be hunky-dory for Ubuntu!

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/amiaopensource/ffmprovisr/pull/181#issuecomment-299334726>,
or mute the thread

privatezero:
@kieranjol uh-oh! I sense more if statements coming!

scripts/ffmprovisr
```diff
@@ -0,0 +1,21 @@
...
+    if [ -d /usr/local/Cellar/ffmprovisr ] ; then
+        ffmprovisr_path=$(find /usr/local/Cellar/ffmprovisr -iname 'index.html' | sort -M | tail -n1)
+    fi
+    if [ -z "${ffmprovisr_path}" ] ; then
+        ffmprovisr_path='https://amiaopensource.github.io/ffmprovisr/'
+    fi
+    open "${ffmprovisr_path}"
+fi
+
+if [[ $OSTYPE = linux-gnu ]] ; then
```
retokromer: Why do you not use an `elif` instruction instead of `fi` `if`? In my opinion, it would be more readable (i.e. easier to maintain): if one OS, then … else if another OS, then …

privatezero: I certainly could change it to use `elif`. With anticipating more conditionals for different systems (like linux on windows) I personally find the symmetry of having individual discrete `if fi` blocks easier to read, but I might be in the minority on that!

retokromer: There is no law! Please do as you prefer. If there is a reason behind, it’s fine for me.

retokromer:
That’s the reason why I use the same – BBEdit which is Unix/Linux inspired – since… half a century on macOS! ;-)

retokromer (approved):

scripts/ffmprovisr
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+if [[ $OSTYPE = darwin* ]] ; then
+    if [ -d /usr/local/Cellar/ffmprovisr ] ; then
+        ffmprovisr_path=$(find /usr/local/Cellar/ffmprovisr -iname 'index.html' | sort -M | tail -n1)
+    fi
```
retokromer, line 6: Hmm… I’m just thinking now if an `else` would not be better: either there is the folder, or the online version is used. Or do I miss something?

scripts/ffmprovisr
```diff
@@ -0,0 +1,19 @@
...
+    if [ -z "${ffmprovisr_path}" ] ; then
+        ffmprovisr_path='https://amiaopensource.github.io/ffmprovisr/'
+    fi
+    open "${ffmprovisr_path}"
+elif [[ $OSTYPE = linux-gnu ]] ; then
+    if [ -d ~/.linuxbrew/Cellar/ffmprovisr ] ; then
+        ffmprovisr_path=$(find ~/.linuxbrew/Cellar/ffmprovisr -iname 'index.html' | sort -M | tail -n1)
+    fi
```
retokromer, line 14: same here

retokromer:
No, an additional test should be added. I am merging your version and making a new release.

privatezero:
Yeah, it felt safer somehow to have the extra test. That way even if the Cellar folder existed, if something went wrong with finding index.html that resulted in a null value it would still try the web source.

ablwr:
Amazing work from the amazing @privatezero !!!!!!!!!!!

retokromer:
Thank you very much, @privatezero, works well!

privatezero:
yay! I have my laptop today and can confirm that it works in Ubuntu 16.04 as well!

retokromer:
@kieranjol Could you possibly test this also on Ubuntu 14?

kfrn:
This looks really good. Thank you for the contributions @privatezero :)

PR #184, opened 2017-05-09
Minor wording amendment

kfrn:
A few small tweaks, plus changing one command from "re-encodes using the same video codec" to "copies the video stream without re-encoding it" (referring to `-c:v copy`).


retokromer (approved):

retokromer:
Thank you, as usual!

PR #185, opened 2017-05-09
add query for default browser

privatezero:
For macOS, adds a check for the default browser and specifies that that ffmprovisr should be opened using that.  Addresses problems found by @todrobbins in https://github.com/amiaopensource/cable-bible/pull/23.

### DIFF
--- a/scripts/ffmprovisr
+++ b/scripts/ffmprovisr
@@ -1,13 +1,18 @@
 #!/usr/bin/env bash
 
 if [[ $OSTYPE = darwin* ]] ; then
+    default_browser=$(plutil -convert json ~/Library/Preferences/com.apple.LaunchServices/com.apple.launchservices.secure.plist -r -o - | grep https -b1 | tail -n1 | cut -d'"' -f4)
     if [ -d /usr/local/Cellar/ffmprovisr ] ; then
         ffmprovisr_path=$(find /usr/local/Cellar/ffmprovisr -iname 'index.html' | sort -M | tail -n1)
     fi
     if [ -z "${ffmprovisr_path}" ] ; then
         ffmprovisr_path='https://amiaopensource.github.io/ffmprovisr/'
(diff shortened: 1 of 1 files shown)


scripts/ffmprovisr
```diff
@@ -1,13 +1,18 @@
 #!/usr/bin/env bash
 
 if [[ $OSTYPE = darwin* ]] ; then
+    default_browser=$(plutil -convert json ~/Library/Preferences/com.apple.LaunchServices/com.apple.launchservices.secure.plist -r -o - | grep https -b1 | tail -n1 | cut -d'"' -f4)
```
retokromer, line 4: I would put this just before…

scripts/ffmprovisr
```diff
@@ -1,13 +1,18 @@
...
     if [ -d /usr/local/Cellar/ffmprovisr ] ; then
         ffmprovisr_path=$(find /usr/local/Cellar/ffmprovisr -iname 'index.html' | sort -M | tail -n1)
     fi
     if [ -z "${ffmprovisr_path}" ] ; then
         ffmprovisr_path='https://amiaopensource.github.io/ffmprovisr/'
     fi
-    open "${ffmprovisr_path}"
+    if [ -n "${default_browser}" ] ; then
```
retokromer, line 11: … that. So the `ffmprovisr_path` part and the `default_browser` part are grouped (easier to read and maintain).

retokromer (approved):

retokromer:
Just a suggestion. LGTM

privatezero:
Hmm, the thought behind assigning it at the top of the subsections was to establish a standard pattern should any other variables need to be assigned (such as a check for ubuntu version in the linux section).  That way readers could look to the top of each `if fi` section and understand where all the values were coming from.

retokromer:
OK

PR #189, opened 2017-05-17
Added command for splitting into segments (segment muxer).

pjotrek-b:
The title says it all :grin: 
Put it right next to "join/concatenate". Thought that makes sense.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1570,19 +1570,29 @@ e.g.: <code>ffmpeg -f concat -safe 0 -i mylist.txt -c copy <i>output_file</i></c
           <h3>Split file into segments</h3>
           <p><code>ffmpeg -i <i>input_file</i> -c copy -map 0 -f segment -segment_time 60 -reset_timestamps 1 <i>output_file-%03d.mkv</i></code></p>
           <dl>
-            <dt>ffmpeg</dt><dd>starts the command</dd>
-            <dt>-i <i>input_file</i></dt><dd>takes in a normal file</dd>
-            <dt>-c copy</dt><dd>use stream copy mode to re-mux instead of re-encode.</dd>
+            <dt>ffmpeg</dt><dd>Starts the command.</dd>
+            <dt>-i <i>input_file</i></dt><dd>Takes in a normal file.</dd>
+            <dt>-c copy</dt><dd>Use stream copy mode to re-mux instead of re-encode.</dd>
             <dt>-map 0</dt><dd>Tells ffmpeg to map all streams of the input to the output.</dd>
-            <dt>-f segment</dt><dd>use <a href="http://www.ffmpeg.org/ffmpeg-formats.html#toc-segment_002c-stream_005fsegment_002c-ssegment">segment muxer</a> for generating the output.</dd>
+            <dt>-f segment</dt><dd>Use <a href="http://www.ffmpeg.org/ffmpeg-formats.html#toc-segment_002c-stream_005fsegment_002c-ssegment">segment muxer</a> for generating the output.</dd>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1560,6 +1560,38 @@
...
+  <div id="segment_file" class="modal fade" tabindex="-1" role="dialog">
+    <div class="modal-dialog modal-lg">
+      <div class="modal-content">
+        <div class="well">
+          <h3>Split file into segments</h3>
+          <p><code>ffmpeg -i <i>input_file</i> -c copy -map 0 -f segment -segment_time 60 -reset_timestamps 1 <i>output_file-%03d.mkv</i></code></p>
+          <dl>
+            <dt>ffmpeg</dt><dd>starts the command</dd>
+            <dt>-i <i>input_file</i></dt><dd>takes in a normal file</dd>
+            <dt>-c copy</dt><dd>use stream copy mode to re-mux instead of re-encode.</dd>
```
retokromer: When you end the sentence with a period, you should start it with a capital.

index.html
```diff
@@ -1560,6 +1560,38 @@
...
+            <dt>-map 0</dt><dd>Tells ffmpeg to map all streams of the input to the output.</dd>
+            <dt>-f segment</dt><dd>use <a href="http://www.ffmpeg.org/ffmpeg-formats.html#toc-segment_002c-stream_005fsegment_002c-ssegment">segment muxer</a> for generating the output.</dd>
```
retokromer: same

index.html
```diff
@@ -1560,6 +1560,38 @@
...
+            <dt>-segment_time 60</dt><dd>Set duration of each segment (in seconds). This example creates segments with max. duration of 60s each.</dd>
+            <dt>-reset_timestamps 1</dt><dd>Reset timestamps of each segment to 0. Meant to ease the playback of the generated segments.</dd>
+            <dt><i>output_file-%03d.mkv</i></dt>
+            <dd>
+            path, name and extension of the output file.<br />
```
retokromer: please use `<br>` (aka HTML 5, instead of XHTML)

index.html
```diff
@@ -1560,6 +1560,38 @@
...
+            In order to have an incrementing number in each segment filename, FFmpeg supports <a href="http://www.cplusplus.com/reference/cstdio/printf/">printf-style</a> syntax for a counter.
+            In this example, '%03d' means: 3-digits, zero-padded<br />
```
retokromer: same

retokromer (changes requested):
Stylistic nitpicking… LGTM!

index.html
pjotrek-b: Roger that.
Got used to XHTML syntax, cause before HTML5 <br/> caused problems in some engines...
Will fix it as you proposed :smile:

retokromer: Thank you! I also personally prefer XHTML, but in practice HTML5 is currently _the_ standard. We moved `ffmprovisr` some months ago.

pjotrek-b:
Fixed the requested changes, one typo and improved printf-digit syntax examples.

retokromer (approved):

retokromer:
Thank you very much indeed!

pjotrek-b:
Thanks! :)

One final question about this:
The segment muxer allows to create a segment-list file that can be used 1:1 for the concat muxer.

Didn't want to bloat the text, but any suggestion/idea where this information might fit in?
(or if it makes sense to mention it at all?)

retokromer:
I guess it makes sense for our audience (as long as I know it… ;-)

PR #202, opened 2017-07-12
convert colourspace - fix ffmpeg link

kieranjol:
Looks like the original was more of a relative path.


retokromer (approved):
I missed that one. Thank you, @kieranjol!

ablwr (approved):

PR #204, opened 2017-07-12
fixes aspect ration typo

kieranjol:
I only just realised that this is the kind of thing that probably doesn't need a pull request, but figured tis best!


privatezero:
Looks good! Good eyes!

retokromer (approved):
Indeed, in my opinion you can merge without PR.

kieranjol:
Aye, thanks both!

PR #205, opened 2017-07-12
change codec explanation for field order

kieranjol:
Pretty sure I wrote this recipe. I think it might be a bit misleading to have `-c:v prores` here in this recipe, as it should instead be a variable. I've seen people use this recipe and use the prores option even if they wanted to maintain a lossless codec.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2083,12 +2083,12 @@ e.g.: <code>ffmpeg -f concat -safe 0 -i mylist.txt -c copy <i>output_file</i></c
       <div class="modal-content">
         <div class="well">
           <h3>Change field order of an interlaced video</h3>
-          <p><code>ffmpeg -i <i>input_file</i> -c:v prores -filter:v setfield=tff <i>output_file</i></code></p>
+          <p><code>ffmpeg -i <i>input_file</i> -c:v <i>video_codec</i> -filter:v setfield=tff <i>output_file</i></code></p>
           <dl>
             <dt>ffmpeg</dt><dd>starts the command</dd>
             <dt>-i <i>input_file</i></dt><dd>path, name and extension of the input file</dd>
             <dt>-filter:v <i>setfield=tff</i></dt><dd>Sets the field order to top field first. Use <code>setfield=bff</code> for bottom field first.</dd>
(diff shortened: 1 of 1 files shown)


retokromer (commented):
I guess we should say for witch codecs this is the case.

kieranjol:
Maybe we could use the command `ffmpeg -codecs | grep 'EV..S '` to show the list of lossless video codecs? Although lots of image only in there too. Or we could just leave out anything about lossnessness and maybe say something along the lines of `enter your desired codec, perhaps the same codec as your input file?`

kfrn:
@retokromer, your typo made me smile so! 


@kieranjol, the change suggested in the commit works for me. I guess you could also mention ffv1?

retokromer:
I second @kfrn suggestion to add `ffv1` to the list.

index.html
```diff
@@ -2083,12 +2083,12 @@
...
         <div class="well">
           <h3>Change field order of an interlaced video</h3>
-          <p><code>ffmpeg -i <i>input_file</i> -c:v prores -filter:v setfield=tff <i>output_file</i></code></p>
+          <p><code>ffmpeg -i <i>input_file</i> -c:v <i>video_codec</i> -filter:v setfield=tff <i>output_file</i></code></p>
           <dl>
             <dt>ffmpeg</dt><dd>starts the command</dd>
             <dt>-i <i>input_file</i></dt><dd>path, name and extension of the input file</dd>
             <dt>-filter:v <i>setfield=tff</i></dt><dd>Sets the field order to top field first. Use <code>setfield=bff</code> for bottom field first.</dd>
-            <dt>-c:v prores</dt><dd>Tells ffmpeg to transcode the video stream into Apple ProRes 422. Experiment with using other codecs.</dd>
+            <dt>-c:v <i>video_codec</i></dt><dd>As a video filter is used, it is not possible to use <code>-c copy</code>. The video must be re-encoded with whatever video codec is chosen, eg. <code>v210</code> or <code>prores</code></dd>
```
retokromer: Nit: should it not be `e.g.`?

retokromer:
Is this OK with you, @kieranjol?

kieranjol:
LGTM! Thanks Reto.

Issue #209, opened 2017-07-15
normalise filename extensions to lower-case


ablwr:
As offered by @kfrn in #207.

PR #210, opened 2017-07-15
Normalise filename extensions (except VOB) to lower-case

kfrn:
Fulfilling #209 
I didn't change .VOB though, since I've never seen .VOB as .vob. But maybe we should opt for complete file extension consistency?


retokromer (approved):
LGTM, thank you!

kieranjol:
LGTM too, thanks!

Issue #213, opened 2017-07-25
Scene Detection using YDIF


drodz11:
In inquiring about using QC Tools for scene detection in a media analysis project, @dericed offered this ffprobe command that (for me, at least) usefully prints a .csv correlating timestamps and their YDIF values. In my tests so far it's a fairly accurate way of determining cuts, but I need to experiment with the thresholds on the spreadsheet side a bit more. Anyway, I figured this might be useful to those (like me) getting acquainted with the functionalities of ffmpeg and already very grateful for ffmprovisr. Code below:

`ffprobe -f lavfi -i movie=YOUR_MOVIE_HERE.MKV,signalstats -show_entries frame=pkt_pts_time:frame_tags=lavfi.signalstats.YDIF -of csv`

Link to original discussion on QC Tools repo: https://github.com/bavc/qctools/issues/422

todrobbins:
🔥 RAD🔥

todrobbins:
`<threadjack>`
How does scene detection using YDIF compare to [scene fingerprinting/perceptual hashing](https://amiaopensource.github.io/ffmprovisr/#compare_video_fingerprints)? I know @privatezero has [worked on this](https://ndsr.americanarchive.org/2017/04/20/adventures-in-perceptual-hashing/) as well, and I'd love to understand these methods better.
`</threadjack>`

drodz11:
@todrobbins ,

Thanks for sharing info re: perceptual hashing! Based on what I've looked at here it seems that the primary difference between scene detection using YDIF vs. perceptual hashing is that large values or "spikes" in YDIF tend to correspond to cuts, whereas perceptual hashing seems more useful in detecting similar content across files using more arbitrary forms of segmentation (i.e. not looking at the media "as edited" but breaking it up into manageable chunks of frames, as discussed by @privatezero ). If you look at the screen grab from QC Tools below, you'll see on the bottom graph a green line representing YDIF values. The "spikes" (most of the time) correlate with cuts in the video. Using the ffprobe command above, I am able to export this YDIF data as a .csv and can then view/manipulate in Excel. Adding a filter to the YDIF column of say >20-25 allows me to see these spikes and their corresponding timestamps. It is not 100% accurate (still experimenting with different thresholds and kinds of video files), but the final result is similar to an EDL you might generate in software like Da Vinci Resolve.

[...]

PR #215, opened 2017-08-07
adds ffprobe command for csv of timecode/ydif #213

ablwr:
#213? 

Oh, I have a question. Is this limited only to specific file format combinations, and should we note that?


retokromer (approved):
LGTM, thank you!

PR #216, opened 2017-08-07
guillemets

retokromer:
- add quotes
- add `target=_blank` to open another window
- fix `</dd>`


ablwr:
Just noticed that too, about the `target=_blank` missing -- but you just caught it!!!

ablwr (approved):

retokromer:
:-)))

Issue #195, opened 2017-06-08
audio phase shift


bturkus:
wondering if anyone else thought it'd be worthwhile to add a phase shift command. I just received a file that was 90 degrees off, and the following command (from superuser.com) seemed to help:

ffmpeg -i input.wav -af "aeval='-val(0)':c=same" output.wav

I'd be happy to add (once I complete eia608!), but I was hoping someone could confirm: is this the right was to do it? And: would it make any sense to add a phase check command? I don't really understand how the one in mmfunctions functions (https://github.com/mediamicroservices/mm/blob/master/phasecheck)

ablwr:
👍

privatezero:
It was 90 degrees as opposed to 180 degrees? I haven't tested `aeval` for phase issues, but if things are a fully reversed you can use the `pan` filter with a negative multiplier to reverse the phase of one of the tracks like such:

`-af pan="stereo| c0=c0| c1=-1*c1"`

This command maps the output's first channel (c0) to the input's first channel and the output's second channel (c1) to the inverse of the input's second channel

privatezero:
I think actually the `*` isn't even required. This is how the phase reverse option is set up in vrecord:
https://github.com/amiaopensource/vrecord/blob/master/vrecord#L382

bturkus:
I struggle with audio in general, but the technician's notes said 90. The orig transfer looked this this:

![screen shot 2017-06-08 at 5 19 58 pm](https://user-images.githubusercontent.com/8987230/26951416-097fb1b2-4c6f-11e7-81c5-40c37cadc75d.png)

And the script I ran turned it to this:

![screen shot 2017-06-08 at 5 41 01 pm](https://user-images.githubusercontent.com/8987230/26952064-c7c9adce-4c71-11e7-98df-34e1567dbdc8.png)

Yours did this:

![screen shot 2017-06-08 at 5 40 37 pm](https://user-images.githubusercontent.com/8987230/26952058-c201acd4-4c71-11e7-8530-3235ba836b7e.png)

I'm super confused at this point, so I'll defer to @privatezero if that's okay. Maybe it's a "half a dozen 
 of one" type thing, though whatever I did looks skinny and weird in QCTools (though it sounded okay). QCTools audio vectorscope is also acting funky (poss a separate issue)

privatezero:
Hi Ben! 
[...]

PR #217, opened 2017-08-07
adds audio phase shift command #195

ablwr:
#195 

CC @bturkus, whom I cannot tag as a reviewer.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1117,7 +1117,7 @@
       <div class="modal-content">
         <div class="well">
           <h3>Flip audio phase shift</h3>
-          <p><code>ffmpeg -i <i>input_file</i> -af pan="stereo| c0=c0| c1=-1*c1" <i>output_file</i></code></p>
+          <p><code>ffmpeg -i <i>input_file</i> -af pan="stereo|c0=c0|c1=-1*c1" <i>output_file</i></code></p>
           <p>This command inverses the audio phase of the second channel by shifting it 180°.</p>
           <dl>
             <dt>ffmpeg</dt><dd>starts the command</dd>
--- a/index.html
+++ b/index.html
@@ -1110,6 +1110,30 @@
   <div class="well">
   <h4>Normalize/Equalize Audio</h4>
 
+  <!-- phase shift -->
+  <span data-toggle="modal" data-target="#phase_shift"><button type="button" class="btn btn-default" data-toggle="tooltip" data-placement="bottom" title="Inverses the audio phase of the second channel">Flip phase shift</button></span>
+  <div id="phase_shift" class="modal fade" tabindex="-1" role="dialog">
+    <div class="modal-dialog modal-lg">
+      <div class="modal-content">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1110,6 +1110,30 @@
...
   <h4>Normalize/Equalize Audio</h4>
 
+  <!-- phase shift -->
+  <span data-toggle="modal" data-target="#phase_shift"><button type="button" class="btn btn-default" data-toggle="tooltip" data-placement="bottom" title="Inverses the audio phase of the second channel">Flip phase shift</button></span>
+  <div id="phase_shift" class="modal fade" tabindex="-1" role="dialog">
+    <div class="modal-dialog modal-lg">
+      <div class="modal-content">
+        <div class="well">
+          <h3>Flip audio phase shift</h3>
+          <p><code>ffmpeg -i <i>input_file</i> -af pan="stereo| c0=c0| c1=-1*c1" <i>output_file</i></code></p>
```
retokromer: Is there a reason for the spacing? I would suggest
```pan="stereo|c0=c0|c1=-1*c1"```
(preferred) or
```pan="stereo | c0=c0 | c1=-1*c1"```
.

File: index.html [hunk 2]
```diff
@@ -1110,6 +1110,30 @@
...
+          <p>This command inverses the audio phase of the second channel by shifting it 180°.</p>
```
retokromer: Does `shifting` also include a rotation in English? If so, then ignore my comment.

index.html
```diff
@@ -1110,6 +1110,30 @@
...
+          <dl>
+            <dt>ffmpeg</dt><dd>starts the command</dd>
+            <dt>-i <i>input file</i></dt><dd>path, name and extension of the input file</dd>
+            <dt>-af</dt><dd>specifies that the next section should be interpreted as an audio filter</dd>
+            <dt>pan=<dt><dd>tell the quoted text below to use the <a href="https://ffmpeg.org/ffmpeg-filters.html#pan-1" target="_blank">pan filter</a></dd>
+            <dt>"stereo| c0=c0| c1=-1*c1"</dt><dd>maps the output's first channel (c0) to the input's first channel and the output's second channel (c1) to the inverse of the input's second channel</dd>
```
retokromer: spacing as above

index.html
ablwr: Just readibility? I cp/pasted from @privatezero in the issue's comments. just updated to the no-spaces which is more standard, I think.

retokromer: Then I would suggest my 2<sup>nd</sup> variant with spaces on both sides of `|`.

index.html
ablwr: rotating is probably better

retokromer (commented):
Despite my nit-picking, LGTM

retokromer (approved):

index.html
retokromer: Sorry, I was mathematician in my previous life.

PR #218, opened 2017-08-07
add sample rate flag to loudnorm

privatezero:
Since loudnorm upsamples to 192 kHz, it seems cleanest/safest to add instructions about specifying output sample rate rather than relying on wrapper defaults and/or having peeps end up with crazy high sample rates in access files!

### DIFF
--- a/index.html
+++ b/index.html
@@ -1166,7 +1166,7 @@
       <div class="modal-content">
         <div class="well">
           <h3>One Pass Loudness Normalization</h3>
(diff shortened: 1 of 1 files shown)


retokromer (approved):
LGTM

ablwr:
👑  @privatezero 👑

ablwr (approved):

kfrn (approved):

Issue #207, opened 2017-07-12
batch processing - mac - add recursive find?


kieranjol:
I think it might be worth adding a `find` example to the batch process recipe for mac? Maybe using the same example as the `for` loop, maybe:
`find -name "*.MXF" -exec ffmpeg -i {} -c copy -map 0 {}.mov \;`
or something?

kfrn:
I like it! Didn't know that method, so thanks. 
What does the `\` do here? I've only used backslashes to continue a command over multiple lines.

(For reference: existing command [here](https://amiaopensource.github.io/ffmprovisr/#batch_processing_bash))

retokromer:
We could also change the extension `.MXF` to `.mxf` in the recipe, because small caps are much more frequent.

dericed:
-iname is case insensitive -name
The {} should be double quoted to avoid spaces breaking things.

retokromer:
>-iname is case insensitive -name

Indeed, but didactically – if we wish to include also beginners – I think it’s better to use the most common writing.

ablwr:
Good to use one and mention the other in the explanation!

kieranjol:
Hi! Thanks all, and yes @dericed , `-iname` would be more useful and I generally use that instead of `-name`. 
@kfrn the slash is just an escaping character for the `;` which ends the statement.

retokromer:
> Good to use one and mention the other in the explanation!

I don’t think that using `.mxf` and mentioning in the explanation `.MXF` would be of interest. And, if really so, should we also double all the `.mov` and `.MOV`, etc.?

kieranjol:
I'm all for using lower case across the board for file extensions..

kfrn:
I'm happy to go through and normalise filename extensions to lower-case.

PR #220, opened 2017-08-15
batch processing mac/linux: adds find command - for #207

kieranjol:
Hi all,

I thought I'd take a stab at this. Looking from the comments in #207, I thought that it might be overkill to do a full, command by command breakdown of `find`, so I just threw in a one-liner. I already did something like this over here so I could just reuse some of this if needed: https://github.com/kieranjol/ifi_batch_workshop#recursive-loop-search-through-all-subfolders

Also, I couldn't figure out how to remove the .mov extension while using `find`. It could be done by piping the output of find to a for loop, but I thought this kept things a lot simpler.

Let me know what ye think anyhow,

### DIFF
--- a/index.html
+++ b/index.html
@@ -1267,7 +1267,7 @@
           <p>The basic pattern will look similar to this:<br>
           <code>for item in *.ext; do ffmpeg -i $item <i>(ffmpeg options here)</i> "${item%.ext}_suffix.ext"</code></p>
           <p>e.g., if an input file is bestmovie002.avi, its output will be bestmovie002_suffix.avi.</p>
-          <p>Variation: recursively process all mxf files in subdirectories using <code>find</code> instead of <code>for</code>:</p>
+          <p>Variation: recursively process all MXF files in subdirectories using <code>find</code> instead of <code>for</code>:</p>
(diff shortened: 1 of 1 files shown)


ablwr (approved):

index.html
```diff
@@ -1267,6 +1267,8 @@
           <p>The basic pattern will look similar to this:<br>
           <code>for item in *.ext; do ffmpeg -i $item <i>(ffmpeg options here)</i> "${item%.ext}_suffix.ext"</code></p>
           <p>e.g., if an input file is bestmovie002.avi, its output will be bestmovie002_suffix.avi.</p>
+          <p>Variation: recursively process all mxf files in subdirectories using <code>find</code> instead of <code>for</code>:</p>
```
retokromer: I would say `all MXF files` with capitals as it’s an abbreviation.

retokromer (approved):
Just one nit-picking, sorry…

index.html
kieranjol: Agreed! I changed it just there. Thanks @retokromer & @ablwr for reviewing.

retokromer (approved):

retokromer:
Thank you, @kieranjol!

ablwr:
always the best!

Issue #233, opened 2017-10-03
Filtergraph explanation


kfrn:
Hi all,  
We now have quite a few commands with relatively complex filtergraphs. I was wondering if it would be worth adding a popup on 'Understanding filtergraphs' at the bottom of ffmprovisr, perhaps under a new heading like 'FFmpeg concepts'.    
I started to write something of this type, incorporating the info included in commands like [IVTC](https://amiaopensource.github.io/ffmprovisr/#inverse-telecine), and some of the content of the [ffmpeg wiki page on filtering](https://trac.ffmpeg.org/wiki/FilteringGuide#FiltergraphChainFilterrelationship).  
The idea is that we could link to this 'info-command' in other commands where appropriate, and maybe even streamline some existing commands.  
What do you all think though? Does it add value, or are things fine as they are?

privatezero:
I think that is a great idea!

retokromer:
Excellent idea!
(FYI: I will make in the next 48 hours or so a new release to use at the FIAF workshop in Chennai.)

kieranjol:
Great idea!

ablwr:
Yes! I think this should maybe go at the top instead of the bottom.

kfrn:
Okay, great! I'll flesh out my draft and get a PR out in the next few days.
@retokromer, I'll not have the chance to do it before you head to Chennai, so go ahead and make that release.

retokromer:
Thank you very much, @kfrn! I made a release with the new structure and design. This way the participants get used to it and have not to change… in the middle of the week. That was an important step forward, in my opinion. I will keep updated the gang about the reactions.

kfrn:
Indeed I thought it would be nice to have the increased navigability for your workshop. Yes, please do let us know what people think.

I shall work on filtergraph etc next week :)

PR #239, opened 2017-10-06
Add basic rewrap command

kfrn:
I'm thinking from the perspective of people new to ffmpeg here. It may be overkill. What do you all reckon?

Next I'll do #233 - not another sidetrack ;-)

### DIFF
--- a/index.html
+++ b/index.html
@@ -64,6 +64,34 @@
   <div class="well">
   <h2 id="rewrap">Change container (rewrap)</h2>
 
+  <!-- Basic rewrap command -->
+  <span data-toggle="modal" data-target="#basic-rewrap"><button type="button" class="btn btn-default" data-toggle="tooltip" data-placement="bottom" title="Basic rewrap command">Basic rewrap command</button></span>
+  <div id="basic-rewrap" class="modal fade" tabindex="-1" role="dialog">
+    <div class="modal-dialog modal-lg">
+      <div class="modal-content">
+        <div class="well">
+          <h3>Rewrap a file</h3>
+          <p><code>ffmpeg -i <i>input_file.ext</i> -c copy -map 0 <i>output_file.ext</i></code></p>
+          <p>This script will rewrap a video file. It will create a new video video file where the inner content (the video, audio, and subtitle data) of the original file is unchanged, but these streams are rehoused within a different container format.</p>
+          <dl>
+            <dt>ffmpeg</dt><dd>starts the command</dd>
+            <dt>-i <i>input_file.ext</i></dt><dd>path and name of the input file<br></dd>
+            <dt>-c copy</dt><dd>copy the streams directly, without re-encoding.</dd>
+            <dt>-map 0</dt><dd>map all streams of the input to the output.<br>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -64,6 +64,34 @@
...
   <h2 id="rewrap">Change container (rewrap)</h2>
 
+  <!-- Basic rewrap command -->
+  <span data-toggle="modal" data-target="#basic-rewrap"><button type="button" class="btn btn-default" data-toggle="tooltip" data-placement="bottom" title="Basic rewrap command">Basic rewrap command</button></span>
+  <div id="basic-rewrap" class="modal fade" tabindex="-1" role="dialog">
+    <div class="modal-dialog modal-lg">
+      <div class="modal-content">
+        <div class="well">
+          <h3>Rewrap a file</h3>
+          <p><code>ffmpeg -i <i>input_file.ext</i> -c copy -map 0 <i>output_file.ext</i></code></p>
```
kieranjol, line 74: +1 for `-map 0`

index.html
```diff
@@ -64,6 +64,34 @@
...
+          <p>This script will rewrap a video file. It will create a new video video file where the inner content (the video, audio, and subtitle data) of the original file is unchanged, but these streams are rehoused within a different container format.</p>
```
kieranjol, line 75: I like the  beginner-friendly terminology here - it's probably for the best to not talk about streams/remuxing etc if the target audience for this recipe is beginners.

index.html
```diff
@@ -64,6 +64,34 @@
...
+            <dt><i>output_file.ext</i></dt><dd>path and name of the output file.<br>
+            The new container you are rewrapping to is defined by the filename extension used here, e.g. .mkv, .mp4, .mov.</dd>
+          </dl>
+          <h4>Important caveat</h4>
+          <p>It may not be possible to rewrap a file's contents to a new container without re-encoding one or more of the streams. Some containers can only contain streams of a certain encoding type: for example, the .mp4 container does not support uncompressed audio tracks. (In practice .mp4 goes hand-in-hand with a H.264-encoded video stream and an AAC-encoded video stream, although other types of video and audio streams are possible). Another example is that the Matroska container does not allow data tracks; see the <a href="#mkv-to-mp4">MKV to MP4 recipe</a>.</p>
```
kieranjol: +1 on explanation

kieranjol (approved):

kieranjol:
LGTM - I love it.

index.html
kfrn: Thanks for the feedback, Kieran. Do you think it's worth adding a note on terminology (i.e. that rewrapping is also called remuxing)?

kieranjol: I think so actually -I've found that any of the muxing terms can kinda scare beginners off a bit as it sounds a lot more technical than it is. Not sure if saying 'remuxing' is prefereable to 're-multiplexing'? Maybe both - 'Rewrapping is also called re-multiplexing/remuxing' ?

kfrn: I've added a new commit which notes this. I take your point, but I do think it's good to mention the different names, for demystification! And so people aren't confused about whether remuxing means something different.

kieranjol: Looks great!

retokromer: It’s a good thing to mention both muxing and wrapping, because people coming from other languages than English sometimes find the one and sometimes the other more… natural.

ablwr:
Reckon you the best

ablwr (approved):

Issue #237, opened 2017-10-04
Refine the organisation of the page


kfrn:
I got the ball rolling with #236, but I think more refinement can be done re: grouping commands into specific categories.

- Are current commands included under the right heading?
- Could commands currently listed under **Other** be moved elsewhere, perhaps under a new category?
   - example: do the commands relating to audio belong under a new heading?
- Are the current headings usefully named?
   - example: could **Change video properties**, which groups commands changing primary video properties like aspect ratio and colourspace, be better named?
- Could some things be better categorised?

etc etc.

@kieranjol expressed some interest in finding homes for some of the currently Misc items.    

If we have any lurkers who have been interested in contributing, this could be a good task for you to look at. 😁

kfrn:
Some comments on this topic came up in #235.

What do you think of the following?

1. Is it worth merging back _**Generate image files from a video**_ & _**Create an animated GIF**_ to a single section, e.g. **Create thumbnails or GIFs**?

2a. What about creating a section called **Work with audio**, into which is collected the following:
- _Fix AV Sync: Resample audio_
- _Extract audio_
- _Combine audio tracks_ 
- _Create access MP3_

2b. and if so, should the current contents of **Normalize/equalize audio** become part of this section, or stay separate?
- _Flip phase shift_
- _Calculate Loudness Levels_
- _RIAA Equalization_
- _One Pass Loudness Normalization_
- _Two Pass Loudness Normalization_

ablwr:
I agree with rolling images and gifs into one -- Create image-based derivatives? Although I prefer the more simple language you offer: "Create thumbnails or GIFs".

Considering the current size/scope of audio-based recipes, I think it's good to lump 'em all together. What do others think?

privatezero:
This all sounds like a good idea to me, with the possible issue that if all the audio was lumped together it might confuse someone who was looking for the WAV -> mp3 and WAV ->AAC scripts that are contained in the Change Codec section.  

Also, in hindsight, looking at _Create access MP3_, that probably should be renamed something more explicit like _Append copyright notice to access mp3__ to differentiate it from the standard WAV -> mp3.

I would be happy to make that change once I get a moment!

retokromer:
I second the idea to group the video codec recipes and the audio codec recipes in two distinct sections.

ablwr:
[...]

PR #240, opened 2017-10-07
Reorganise some recipes + add ffmpeg basics & filtergraph command

kfrn:
- Filtergraph explanation added in new section, **FFmpeg concepts**
- New section, **Create an animated GIF**, with relevant recipes moved there
- _Modify speed_ moved to **Change video properties**, since I consider speed a pretty fundamental video property.

This addresses #233, and some of the things I mentioned in #237.

### DIFF
--- a/index.html
+++ b/index.html
@@ -43,6 +43,7 @@
   <div class="well">
     <h3>What do you want to do?</h3>
     <p class="select-from">Click one of the following categories to see possible commands of that type:</p>
+    <a href="#concepts"><button type="button" class="btn contents-list">Understand ffmpeg concepts</button></a>
     <a href="#rewrap"><button type="button" class="btn contents-list">Change container (rewrap)</button></a>
     <a href="#transcode"><button type="button" class="btn contents-list">Change codec (transcode)</button></a>
     <a href="#properties"><button type="button" class="btn contents-list">Change video properties</button></a>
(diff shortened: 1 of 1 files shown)


retokromer:
The problem I see is that now at the beginning are the _FFmpeg concepts_ which starts with the _Filtergraph_ explanation, a pretty advanced topic. In my opinion, didactically speaking, the starting should be much gentler (e.g. on [FFmpeg’s website](https://ffmpeg.org/) they start with `ffmpeg -i input.mp4 output.avi`).

kfrn:
That's a good point, @retokromer. Do we call this **Advanced/Further concepts** instead and move the section to the bottom of the page? Or keep it up top, adding before the _Filtergraph_ explanation a _Basic structure of an ffmpeg command_ entry? I certainly don't want to scare away any people new to ffmpeg.

retokromer (commented):
I suggest not to start with the filtergraphs, but to move this down.

retokromer:
I guess _advanced concepts_ should be fine. But I would not start with this.

retokromer:
Adding the _basic structure_ at the beginning is excellent indeed! Yet I would still separate between basic and advanced.

index.html
```diff
@@ -62,6 +64,41 @@
...
+          <p>The use of a filter is signalled by the flag <code>-vf</code> (video filter) or <code>-af</code> (audio filter), followed by the name and options of the filter itself. For example, take the <a href="#convert-colourspace">convert colourspace</a> command:</p>
+          <p><code>ffmpeg -i <i>input_file</i> -c:v libx264 -vf colormatrix=<i>src</i>:<i>dst</i> <i>output_file</i></code>
+          <p>Here, <a href="http://ffmpeg.org/ffmpeg-filters.html#colormatrix">colormatrix</a> is the filter used, with <i>src</i> and <i>dst</i> representing the source and destination colourspaces. This part following the <code>-vf</code> is a <b>filtergraph</b>.</p>
+          <p>It is also possible to apply multiple filters to an input, which are sequenced together in the filtergraph. A chained set of filters is called a filter chain, and a filtergraph may include multiple filter chains. Filters in a filterchain are separated from each other by commas (<code>,</code>), and filterchains are separated from each other by semicolons (<code>;</code>). For example, take the <a href="#inverse-telecine">inverse telecine</a> command:</p>
```
kieranjol: The filtergraph section looks great to me, very helpful and easy to understand. My only comment here is that the order of filters is also significant (i think) - for example, it seems to be recommended that interlaced material is deinterlaced prior to a rescale.

kfrn:
I took your advice, @retokromer - we now have basic and advanced sections.  
I found the writing the _Basic structure_ explanation a bit tricky, terminology-wise, so I welcome any suggestions on that.

index.html
```diff
@@ -42,26 +42,60 @@
...
+          <p><b>Note:</b> flags are always prepended with a hyphen.</p>
+          <p>A very basic ffmpeg command looks like this:</p>
+          <dl>
+            <dt>ffmpeg</dt><dd>starts the command</dd>
+            <dt>-i <i>input_file.ext</i></dt><dd>path and name of the input file<br></dd>
+            <dt><i>-flag some_action</i></dt><dd>tell ffmpeg to do something, by supplying a valid flag and action</dd>
+            <dt><i>output_file.ext</i></dt><dd>path and name of the output file.<br>
+              Because this is the last part of the command, the filename you type here does not have a flag designating it as the output file.</dd>
+          </dl>
+
```
kieranjol: is this space needed?

index.html
```diff
@@ -42,26 +42,60 @@
...
   <div class="well">
+  <h2 id="getting-started">Getting started with ffmpeg</h2>
+
+  <!-- Basic structure of an ffmpeg command -->
+  <span data-toggle="modal" data-target="#basic-structure"><button type="button" class="btn btn-default" data-toggle="tooltip" data-placement="bottom" title="Basic structure of an ffmpeg command">Basic structure of an ffmpeg command</button></span>
+  <div id="basic-structure" class="modal fade" tabindex="-1" role="dialog">
+    <div class="modal-dialog modal-lg">
+      <div class="modal-content">
+        <div class="well">
+          <h3>Basic structure of an ffmpeg command</h3>
```
kieranjol: nit: for consistency - I think we were adhering to the command line argument being lowercase ffmpeg, the software itself as FFmpeg?

index.html
```diff
@@ -42,26 +42,60 @@
...
+          <p>At its basis, an ffmpeg command is relatively simple. After you have installed ffmpeg (see instructions <a href="https://avpres.net/FFmpeg/#ch1">here</a>), the program is invoked simply by typing <code>ffmpeg</code> at the command prompt.</p>
```
kieranjol: another FFmpeg - i won't make any more comments about this as i don't want to clog the feed.

index.html
```diff
@@ -42,26 +42,60 @@
...
+          <p>Subsequently, each instruction that you supply to ffmpeg is actually a pair: a flag, which designates the <i>type</i> of action you want to carry out; and then the specific action itself.</p>
+          <p>For example, with the instruction <code>-i <i>input_file.ext</i></code>, the <code>-i</code> flag tells ffmpeg that you are supplying an input file, and <code>input_file.ext</code> states which file it is.</p>
+          <p><b>Note:</b> flags are always prepended with a hyphen.</p>
+          <p>A very basic ffmpeg command looks like this:</p>
+          <dl>
+            <dt>ffmpeg</dt><dd>starts the command</dd>
+            <dt>-i <i>input_file.ext</i></dt><dd>path and name of the input file<br></dd>
+            <dt><i>-flag some_action</i></dt><dd>tell ffmpeg to do something, by supplying a valid flag and action</dd>
```
kieranjol: I was wondering if an example could be used here like -c:v prores, but then we'd have to explain that too, so maybe this is all just perfect as is.

kieranjol:
I added a few comments, this is yet another great contribution Katherine.

kfrn:
I can't see them, @kieranjol - maybe you did not submit the review? I've been caught out by that before.

kieranjol:
Aye,sorry Katherinem Should be OK now?

index.html
kfrn: Oh good catch, I had meant to say something about filter order being important. Fixing this!

index.html
kfrn: This is inconsistent throughout, so in my new commit I've tried to harmonise across the whole app.

index.html
kfrn: I did write that exact example, but then took it out - readding!

kfrn:
Thanks heaps for the feedback @kieranjol, I've made those changes, plus a couple of other things that I caught.

I'm actually not in favour of using <i>target="_blank"</i>, but we might as well be consistent.

ablwr:
Wowwwwwwwwwwwww, what a treat to wake up to!

ablwr:
Use of "Getting started with FFmpeg" and "Learn about the basics of FFmpeg" is inconsistent -- better to use one or the other. I would also put the advanced concepts section under the Basics one, grouping them semantically by their larger theme, rather than having the advanced one get lost at the bottom.

ablwr:
Katherine, this is really really amazing!

ablwr (approved):

index.html
retokromer: ProRes is a little tricky, because Apple stopped supporting QuickTime on Window, and an archive should use Matroska instead. (More on this in Vienna.)

retokromer:
I started using `target="_blank"` for external links, in order to have still the original window. If you don’t wish this behaviour, please remove them.

kfrn:
@ablwr - agreed and done! 

@retokromer I'm one of those people who thinks that <i>target="_blank"</i> is bad UX. But it's a minor thing, I'm not pushing for a wholesale change.

kfrn:
Mergin' my own pull requests, yeeeeaaaaaaa

PR #241, opened 2017-10-09
removes junk that accidentally came into ffmprovisr

ablwr:
Accidentally brought in this trash when I changed the headers -- no idea how that happened other than I wasn't paying attention. Anyway, it can go.


kfrn:
Cheers for picking this up, I'd assumed it was there intentionally!

retokromer:
Me too, I never got into those «details», sorry. Thank you, @ablwr, for cleaning up!

Issue #235, opened 2017-10-03
r e - d e s i g n ❓


ablwr:
Thinking about the open and recently closed issues of header tags #228, filtergraph explanation #233, search functionality #211, and doing a lot of personal-site refactoring to reduce code bloat and increase responsiveness/accessibility, I am thinking...

# Is it time for a re-design?  

I'm very happy to take on this task, but I need some requirements (constraints) from the community to think about the best ways to make ffmprovsr better visually. I think I heard @retokromer mention that students found it visually intimidating (because of the black? because of all the options at once?).

Fellow maintainers, contributors, interested parties, what do you think?

kieranjol:
I think a re-design would be very helpful - I keep hearing from users that it is difficult to find what you want on the page - though they are then full of praise for what ffmprovisr teaches them..

I kinda like the black btw..

And perhaps some refactoring is needed so that things like 'ffmpeg concepts' could be incorporated without adding a flood of more items. I love the idea of the 'concepts', there's room for a lot of tips and tricks that might be worthy of a dedicated ffmpeg wiki page.

retokromer:
>because of the black?

No.
>because of all the options at once?

Yes. And without a clear hierarchy and/or workflow. They need some time to find where to start – today, people are impatient… ;-)

ablwr:
I like the black background too because that is what command line interfaces mostly default to. :)

ablwr:
Not thrilled about our use of modals either -- I think it'd be better to have the text grow/expand out of the button but stay on the page rather than pop up on top of it.

From the technical side, we are also using jQuery and Bootstrap but really don't need either, so we can eliminate a lot of code this way and simplify things. We can probably simplify the formula with which we add new scripts, too.

retokromer:
)Side note 1: currently, ffmprovisr does not work fine on a Braille terminal :-(
(Side note 2: we should say how to pronounce ffmprovisr ;-)

kfrn:
I'm not in love with modals either - a 'click to expand' approach would be good, I think.

ablwr:
I'll be thinking about this. Please ask around or relay what else you've learned from other people coming to this for the first time! I am going to tweet about it cuz that seems to get good results.

ablwr:
I LOVE ASKING PEOPLE WHAT THEY DISLIKE

ablwr:
[...]

PR #243, opened 2017-10-09
initial redesign: grow ffmprovsr

ablwr:
Related to #235, this puts the layout in a grid (and relational-sized instead of using pixels) instead of relying on Bootstrap, which will help in the future and sets a media size breakpoint so smaller screens see the new (thanks @kfrn!) table of contents at the top of a single column, but bigger screens push it to the sidebar on the left. Adds meta declaration for viewport width assessment.

I want to do more but I want to keep PRs small.

### DIFF
--- a/css/css.css
+++ b/css/css.css
@@ -1,9 +1,74 @@
+.grid {
+  display: grid;
+  align-content: center;
+  grid-template-areas:
+   "header"
+   "sidebar"
+   "content"
+   "footer"
+}
+
+.header {
+  grid-area: header;
+  text-align: center;
+
+}
+
+.content {
+  grid-area: content;
+}
+
+.sidebar {
+  grid-area: sidebar;
+}
+
+.footer {
+  grid-area: footer;
+  text-align: center;
+}
+
+@media only screen and (min-width: 1px)  {
+.grid {
+  grid-template-columns: 1fr;
+  grid-template-areas:
+  "header"
+  "sidebar"
+  "content"
+  "footer";
+}
+}
+
+@media only screen and (min-width: 1000px)   {
+  .grid {
+    grid-gap: 1rem;
+    grid-template-columns: 0.25fr auto;
+    grid-template-areas:
+    "header  header"
+    "sidebar content"
+    "footer  footer";
+  }
+}
+
+.license img {
+  max-width: 10em;
+  display: inline;
+}
+
 h1, h2, h3, h4, h5, h6 {
   font-family: 'Montserrat', sans-serif;
 }
 
+h1 {
+  letter-spacing: 0.1em;
+  font-size: 7vw;
+}
+
 h2 {
-  font-size: 24px;
+  font-size: 3em;
+}
+
+h3 {
+  font-size: 2em;
 }
 
 .intro-lead {
@@ -39,11 +104,6 @@ img {
   margin-bottom: 18px;
 }
 
-h1 {
-  letter-spacing: 8px;
-  font-size: 86px;
-  line-height: 120px;
-}
 
 div {
   font-family: 'Merriweather', serif;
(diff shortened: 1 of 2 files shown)


css/css.css
```diff
@@ -1,9 +1,74 @@
...
+   "header"
+   "sidebar"
+   "content"
+   "footer"
+}
+
+.header {
+  grid-area: header;
+  text-align: center;
+
```
retokromer: blank line not needed

css/css.css
```diff
@@ -1,9 +1,74 @@
+.grid {
+  display: grid;
+  align-content: center;
+  grid-template-areas:
+   "header"
+   "sidebar"
+   "content"
+   "footer"
```
retokromer: Nit: I guess (not verified), we have a `;` also at the end of the last item.

css/css.css
```diff
@@ -1,9 +1,74 @@
...
+
+@media only screen and (min-width: 1px)  {
+.grid {
+  grid-template-columns: 1fr;
+  grid-template-areas:
+  "header"
+  "sidebar"
+  "content"
+  "footer";
+}
```
retokromer: I suggest to indent this level.

ablwr:
I wouldn't get too caught up with perfecting this code since I anticipate it changing, what's more important is if the style/resizing is acceptable.

ablwr:
I still made the changes though. 😅

index.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <title>ffmprovisr</title>
-    <meta charset="utf-8">
+    <meta name="viewport" content="text/html, charset=UTF-8, width=device-width, initial-scale=1">
```
retokromer: Is this widely supported? I mean: can the `<meta charset="utf-8">` be grouped into this in HTML5?

ablwr: Yeah. Mostly this sets the default sizing so they can change as needed to fit onto different-sized devices. In this branch, only the h1 tag changes size based on how big the window is so it doesn't get smushed.

retokromer:
I just saw it, going through the code…

index.html
retokromer: Hmm… not sure. On Safari it gives now:

bildschirmfoto 2017-10-09 um 15 40 35

retokromer: It sets the standard charset of the browser, and does not understand to change. Putting again back that line, gives:

bildschirmfoto 2017-10-09 um 15 42 47

ablwr: Arg, thanks, hmm... I mean the charset=utf-8 *is* still in there. It might need an apostrophe instead of a comma. One sec!

retokromer (commented):
Except the charset issue, it works fine on my side.

index.html
ablwr: Sorry about that, I totally missed that charset="" was not scoped within the content=""!

retokromer (approved):
LGTM (tested only on my laptop)

ablwr:
I will wait for @kfrn, who has a lot of CSS knowledge. Then I will work on tackling the bigger issue of replacing modals with inline...

retokromer:
I suggest, once @kfrn has agreed, to merge this PR, and work on a new one.

kfrn:
"A lot of CSS knowledge" lol. I shall have a look when I'm at home this evening.

retokromer:
BTW: Considering the reaction of my 1st practical class yesterday (out of 8 this week), the changes already made go in the right direction. Thanks a lot!

kfrn:
It looks great @ablwr ! Thanks so much for doing this!

kfrn (approved):

kfrn:
Also - I'm really glad to hear that, @retokromer !

ablwr:
Thanks so much to @kfrn!!!!!

retokromer:
Thank you @ablwr and @kfrn for the clean-up! I just made a new release.

PR #248, opened 2017-10-13
modals go inline instead of popup

ablwr:
This is a major re-write to the scripts, changing the structure of the script template. It moves from the current version of "popping up" a modal to having clicking open up the recipe in-line. Also, multiple recipes can be open at the same time.

I think there is some more styling to do but wanted to open this PR first before I make more changes. Also, the Javascript that generates the link-to-this-recipe can be cleaned up or possibly eliminated, but I set it up to be functional at this time so we can focus on if this is a desired change or not, and I can move forward from there.

newffmprovsr

### DIFF
--- a/index.html
+++ b/index.html
@@ -21,6 +21,7 @@
 
   <nav class="well sidebar">
     <h3>Table of Contents</h3>
+    <p class="link"></p>
     <p class="select-from">Click one of the following categories to see commands of that type:</p>
     <a href="#about"><button type="button" class="btn contents-list">About this resource</button></a>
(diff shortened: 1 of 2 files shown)


retokromer (approved):
LGTM (I didn’t test all, but what we just tested IRL worked fine.)

kieranjol:
It looks great just based on the gif example,will look properly later when
at a computer.

kfrn:
It looks great, Ashley! 👍 Awesome work 💚

kfrn (approved):

PR #256, opened 2017-10-14
turns ToC buttons into div link table

ablwr:
Makes the Table of Content not (html) buttons anymore, instead a series of wide clickable links that change color on hover. Also I changed the color of links (when not the ToC) from the default blue to GREEN in an homage to the ffmpeg homepage.

### DIFF
--- a/css/css.css
+++ b/css/css.css
@@ -1,3 +1,7 @@
+html, body {
+  font-size: 16px;
+}
+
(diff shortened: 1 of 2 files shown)


kfrn:
LGTM Ashley!

retokromer:
LGTM

PR #258, opened 2017-10-14
Moves CSS around, fixes overflow error on small devices

ablwr:
moves visual style to "local" css (shows how little of bootstrap we are using for style), simplifies button style to one class, breaks words on small screens

### DIFF
--- a/css/css.css
+++ b/css/css.css
@@ -1,5 +1,10 @@
 html, body {
   font-size: 16px;
+  line-height: 1.5;
+  color: #888888;
+  background-color: #060606;
+  margin: 0;
+  word-break: break-word;
 }
 
 .grid {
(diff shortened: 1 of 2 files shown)


kfrn:
LGTM! Out of curiosity: why are the explicit inherits for `button` necessary?

ablwr:
They are necessary because other parts of bootstrap wants to overwrite them, I think. If I manage to ditch the bootstrap css, I can ditch this extra code too. It took a while to figure out this is what was causing the problem!

kfrn:
Right, gotcha! I've used Bootstrap, but not extensively. At work we're using a newish framework called Bulma, which is very nice so far.

ablwr:
There's nothing wrong with it although people like to give it shit, I am just obsessed with minimizing dependencies and removing all the stuff we don't use or need LIKE A BIG NERD

PR #259, opened 2017-10-15
Styling changes

kfrn:
I've changed the spacing on the `<h2>`s to line up with the buttons, as well as making the headings a bit smaller.
before
headings-before
after
headings-after

Background of ToC a darker grey, nicer on the eye:
before 
toc-before
after
toc-after

Media query for viewing on a large screen. On my desktop computer, I find it a little much when the page is totally full.
before
desktop-before
after
desktop-after

Most of this stuff is a matter of taste, so I'm at ease if you don't want to use some of the changes.

### DIFF
--- a/css/css.css
+++ b/css/css.css
@@ -94,7 +94,7 @@ h2 {
   margin: 5px 8px 8px 8px;
 }
 
-#about {
+.heading {
   font-size: 2em;
   margin: 6px 0px 12px 0px;
 }
--- a/css/css.css
+++ b/css/css.css
@@ -1,5 +1,4 @@
 html, body {
-  font-size: 16px;
   line-height: 1.5;
   color: #888888;
   background-color: #060606;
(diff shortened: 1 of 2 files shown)


retokromer (commented):
- The higher contrast between background and writing is easier to read for impaired people. That’s a good catch, in my opinion.
- The same font size for «Table of Contents» and «About ffmprovisr» would be simpler.
- The bigger border on the right and the left side seem to me too big.

kfrn:
Thx for the feedback @retokromer 👍 

Now using the same styling for the ToC and About ffmprovisr headings.

If it makes a difference, the media query that adds the spaces at the left and right doesn't kick in until the screen is > 1500 px wide. The screenshot is from my desktop screen (1920 x 1080).

retokromer (approved):

ablwr:
Thank you yes!!!!!!!!!!!!!!!!!!!!!!

ablwr (approved):

ablwr (approved):

PR #262, opened 2017-10-18
changes buttons into div-rows

ablwr:
Like this?

screen shot 2017-10-17 at 21 00 40

screen shot 2017-10-17 at 21 02 27

Restructured, got rid of tooltips, lengthened recipe names if necessary, got rid of <span> wrapping buttons, replaced buttons with divs, renamed btn class to recipe, added green hover effect, made VHS/comet grow in a better way.

### DIFF
--- a/css/css.css
+++ b/css/css.css
@@ -183,6 +183,7 @@ div {
   font-size: 1em;
   border: 1px solid black;
   transition: all .5s ease-out;
+  position: relative;
 }
 
 .contents-list:hover {
--- a/css/css.css
+++ b/css/css.css
@@ -132,88 +132,75 @@ img {
   margin-bottom: 18px;
 }
 
-
 div {
   font-family: 'Merriweather', serif;
   color: white;
 }
 
-button {
-  font: inherit;
-  font-family: inherit;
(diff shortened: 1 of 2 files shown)


retokromer:
Is the colour of the text in the green fields #000000? (Just asking, because the contrast between background and text seems to me to be lower.)

ablwr:
Hmm... using "named" colors of white and black (except for the green).

ablwr:
I boosted the size slightly and switched to the sans-serif font, maybe that is causing the difference?

retokromer:
It’s a matter of contrast: green to black is less than dark grey to white. I just noticed it, and this may make reading more difficult for view impaired people. Yet this is not blocking. (I don’t think that a lighter green would be nice.)

ablwr:
No, contrast is important. I thought you meant it looked different from before. I'll take a look at it and refine so we get better visibility/contrast (although not tonight, but tomorrow).

retokromer (approved):
Otherwise, LGTM.

ablwr:
I'll work on contrast in a new branch!

PR #263, opened 2017-10-18
cleans up hover css, keeps text white on hover

ablwr:
Decided to keep the text white when changing to green -- changing text to black was a mistake! This should improve contrast issues. Also cleaned up the code to combine them, since they were very similar in structure with only a few changes in terms of resizing the icons on hover.

### DIFF
--- a/css/css.css
+++ b/css/css.css
@@ -137,70 +137,47 @@ div {
   color: white;
 }
 
-.recipe {
+.sidebar a {
+  text-decoration: none;
+}
+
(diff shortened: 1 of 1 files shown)


retokromer (approved):
Yes, I guess that way it’s working well. Thank you very much, @ablwr!

ablwr:
I don't know how I missed that here on Github and on the ffmpeg.org site, it's still white-on-green and not black-on-green. The filesize is a bit bigger but hopefully not too big.

PR #265, opened 2017-10-18
removes reliance on bootstrap; moves necessary code to css/js

ablwr:
#235 

*deep breathe* 

Removes Bootstrap dependencies, moves required code to css and js files, changes collapse code to simple CSS and replaces divs with input/checkbox method, wraps ToC into a "well" class to align with content, indents code.

I can also work to remove jQuery pretty easily but I'll do that later and that will BE THE LAST THING.


ablwr:
I think I'm done tinkering with this for now.

kfrn:
Phew!
I'll have a look at home tonight :)

retokromer:
I would put the `Link to this command` at the end of the commands, not at the beginning.

retokromer:
I guess `Flip video image` could be moved to `Change video properties`.

retokromer (approved):
The style change LGTM. See also remarks above.

ablwr:
Oh good point. I had moved it to the top when we had the buttons because there wasn't really a clear distinction when opening, that it was not related to something else. I can update these requests while I wait for @kfrn's review. It is a big change!

index.html
```diff
@@ -370,10 +367,11 @@ <h2 id="transcode">Change codec (transcode)</h2>
...
       </dl>
       <p>A couple notes</p>
-        <ul>
-          <li>About ID3v2.3 tag: ID3v2.3 is better supported than ID3v2.4, FFmpeg's default ID3v2 setting.</li>
-          <li>About dither methods: FFmpeg comes with a variety of dither algorithms, outlined in the <a href="https://ffmpeg.org/ffmpeg-resampler.html" target="_blank">official docs</a>, though some may lead to unintended, drastic digital clipping on some systems.</li>
-        </ul>
+      <ul>
+        <li>About ID3v2.3 tag: ID3v2.3 is better supported than ID3v2.4, FFmpeg's default ID3v2 setting.</li>
+        <li>About dither methods: FFmpeg comes with a variety of dither algorithms, outlined in the <a href="https://ffmpeg.org/ffmpeg-resampler.html" target="_blank">official docs</a>, though some may lead to unintended, drastic digital clipping on some systems.</li>
+      </ul>
```
retokromer: 👍 alignment

retokromer (approved):

index.html
ablwr: Looking at alignment at the very bottom of the doc right now... agggghhhh! There is an indention off somewhere...

ablwr: Found it, haha.

retokromer:
If it’s not too complicated, I would suggest to have the absolute path to the website, instead of the relative, also when used locally:

bildschirmfoto 2017-10-19 um 14 25 48

index.html
retokromer: :-)))

ablwr:
Is that not the absolute path? This works by simply grabbing whatever is displaying in the URL bar at the time.

ablwr:
Do you mean you want it to always display the website path instead of the local file:///Path/to/where/you/put/thisRepo?

ablwr:
Either way, please make a separate issue to work on after this one is complete; I want to pull out the jQuery and replace our sparse Javascript (which causes this to exist) and maybe not even use Javascript at all to put the links there and only use it to update the anchor tag in the URL.

bturkus:
super small thing, and not sure if this is the right place to note it, but I noticed that if you click on "about this resource" at the top of the TOC, you don't jump down automatically.
looks awesome!

ablwr:
Fixed! Thanks @bturkus

retokromer:
@ablwr I was thinking about
```
https://amiaopensource.github.io/ffmprovisr/index.html#flip_image
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

(This way, when printed, it gives the online address.)

retokromer (approved):

ablwr:
I'll make an Issue @retokromer and tackle it there!

ablwr:
I am going to merge this since it fixes the things that rely on javascript -- fundamentally, linking -- which currently is not working on `master`, although I encourage more feedback and review.

PR #272, opened 2017-10-30
Fix sizing issue at large resolutions

kfrn:
So, I noticed that the `min-width: 1500px` media query wasn't working correctly. It seemed to be kicking in at 1200px instead of 1500px. This commit fixes this. To explain visually:

BEFORE --

Here I'm at just under 1200px. Full screen coverage, as you'd expect.
ffmprovisr_mq_1

Then I widen to just over 1200px ... and the 1500px `min-width` media query kicks in! Qué?
ffmprovisr_mq_2

AFTER -- 

At widths up til 1500px, the media query doesn't kick in.
ffmprovisr_mq_3

Then it does!
ffmprovisr_mq_4

I'm not totally sure exactly why the bug was occurring - I'm not well versed in CSS grid yet. So I'll appreciate your eagle eyes, @ablwr 👀 😄

### DIFF
--- a/css/css.css
+++ b/css/css.css
@@ -61,6 +61,10 @@ html, body {
   .grid {
     max-width: 70%;
     margin: 0 auto;
+    grid-template-areas:
+    "header  header"
+    "sidebar content"
+    "footer  footer";
   }
 }
 


retokromer (approved):
LGTM (tested only on two devices/browsers)

ablwr:
Oh interesting! Yeah this makes sense to me, kinda, it was inheriting the default, which is the one-row. I think in a perfect world, there would be no media-queries. I'm super sleepy but I accept this improvement, thank you @kfrn !

PR #273, opened 2017-10-31
Add max width on example images

kfrn:
In the distant past, there was a `max-width` property set on the `img` element, but this has been lost in the shuffle. I've added something similar back, so that the images aren't super giant.

Before -
ffmprovisr_before

After -
ffmprov_after

It may still need tweaking.

### DIFF
--- a/css/css.css
+++ b/css/css.css
@@ -138,7 +138,10 @@ img {
 }
 
 .sample-image {
+  margin: 0 auto;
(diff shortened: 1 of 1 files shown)


retokromer (approved):
LGTM

ablwr (approved):

ablwr:
YAYAYAYAYAY

PR #274, opened 2017-11-01
adds likebreaks in very long codes for firefox support

### LINKED ISSUE #270: Sizing problem in Firefox
I usually use Chrome, but opened FF the other day and found something off with the sizing:

This is full-screen in Firefox (`56.0.1 (64-bit)`). Note the X-scroll:
firefox_fullscreen
As the spacing indicates, even when scrolled all the way right, there's not room for the side spacer.

I had a fiddle with the media queries, thinking it might be something to do with [this](https://github.com/amiaopensource/ffmprovisr/blob/fe937d7e4a8f9482ee2a5e44b6392c722183a43e/css/css.css#L60).  
Toggling off `max-width: 70%;` improved things a little, but didn't fix the problem. It looks like this is happening at the `body` level, anyway.

With `max-width: 70%;` turned off:
firefox_fullscreen_mq

For comparison, here's how it looks in Chrome (`62.0.3202.62`):
chrome_fullscreen

@ablwr, can you offer your CSS expertise?

ablwr:
This is to support ticket #270. Firefox does not elegantly support word-break, which causes the overflow to continue to grow. I couldn't figure out a good CSS solution around this -- there probably is one, though -- so I forced some breaks in the HTML, which lets Firefox size itself as it should.

But is this confusing for beginners copy-and-pasting? It may cause spaces to be added when they shouldn't... I want to test this more but I'm submitting the PR now so I can move on to other work.

resolves #270

### DIFF
--- a/index.html
+++ b/index.html
@@ -793,7 +793,8 @@
     <input type="checkbox" id="loudnorm_two_pass">
     <div class="hiding">
       <h3>Two Pass Loudness Normalization</h3>
-      <p><code>ffmpeg -i <i>input_file</i> -af loudnorm=dual_mono=true:measured_I=<i>input_i</i>:measured_TP=<i>input_tp</i>:measured_LRA=<i>input_lra</i>:measured_thresh=<i>input_thresh</i>:offset=<i>target_offset</i>:linear=true -ar 48k <i>output_file</i></code></p>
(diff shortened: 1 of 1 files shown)


ablwr:
Wait, I got it!

PR #275, opened 2017-11-01
breaks long long lines for code and dt blocks

### LINKED ISSUE #270: Sizing problem in Firefox

ablwr:
Unlike #274, resolves #270 by breaking up the very stubborn code blocks which hurt Firefox!

### DIFF
--- a/css/css.css
+++ b/css/css.css
@@ -120,6 +120,8 @@ code {
   border-radius: 4px;
   word-break: break-all;
   word-wrap: break-word;
+  white-space: normal;
+  display: inline-block;
 }
 
 dd {
@@ -128,6 +130,10 @@ dd {
 
 dt {
   font-family: Menlo, Monaco, Consolas, "Courier New", monospace;
+  word-break: break-all;
+  word-wrap: break-word;
+  white-space: normal;
+  display: inline-block;
 }
 
 img {


css/css.css
```diff
@@ -120,6 +120,8 @@ code {
   border-radius: 4px;
   word-break: break-all;
   word-wrap: break-word;
+  white-space: normal;
```
retokromer, line 123: Is `normal` not the default value of `white-space`?

ablwr:
I think the `<code>` blocks are behaving differently (on Firefox only) so we must be explicit here.

retokromer:
I’ve just checked the specs. It’s indeed the default. Firefox is buggy.

retokromer (approved):

ablwr:
I don't disagree!

Issue #276, opened 2017-11-01
Avoid hyphenation of the command


retokromer:
Is there a simple way in the CSS to avoid «hyphenation» of the command in the middle of a word, like `cop` and `y` as in the following example?

<img width="863" alt="bildschirmfoto 2017-11-01 um 16 47 39" src="https://user-images.githubusercontent.com/3246607/32283442-9bf07e14-bf24-11e7-9706-389f8645a300.png">

cc @ablwr @kfrn

ablwr:
I can keep investigating the issue but that was what I just put into place to keep the code example from going over the page wide and messing up view in Firefox!

retokromer:
FYI: I was using Safari here.

ablwr:
Right, the line breaks happen in all browsers. The lack of linebreak overflow was previously causing the problems in Firefox.

retokromer:
@ablwr Would adding`hyphens: none;` resolve this?
```
code {
  padding: 2px 4px;
  color: #c7254e;
  background-color: #f9f2f4;
  border-radius: 4px;
  word-break: break-all;
  word-wrap: break-word;
  white-space: normal;
  display: inline-block;
  hyphens: none;
}
```

todrobbins:
`line-break: strict;` in the CSS should fix most cases of this, although Firefox is yet to implement... See: https://developer.mozilla.org/en-US/docs/Web/CSS/line-break

retokromer:
Thank you, @todrobbins! I suggest that we move forward that way, and ignore Firefox.

ablwr:
Firefox should not be left behind!

retokromer:
I am unhappy with the idea of having to manage different CSS behaviours, because of the inconsistency of the current browsers.

ablwr:
It breaks on `copy` because it is necessary, otherwise lines do not know how to break with something lengthy like `"[0:a:0]asplit=2[a][b];[b]afifo[bb];[1:a:0][bb]concat=n=2:v=0:a=1[concatout]"`.

ablwr:
An example of why this linebreak has to happen is not for `copy` but for `loudnorm=dual_mono=true:measured_I=input_i:measured_TP=input_tp:measured_LRA=input_lra:measured_thresh=input_thresh:offset=target_offset:linear=true` within https://amiaopensource.github.io/ffmprovisr/#loudnorm_two_pass

ablwr:
Ah! I have a compromise. One moment.

[...]

PR #281, opened 2017-11-28
lifts the break-all and sets a max-width on code for long lines

ablwr:
This is for #276, please test on many browsers and systems. Sets a maximum size for the code block and breaks at the line when necessary. When screens are very small, it will break at any point it deems appropriate.

Works on Mac with Firefox (my default), Chrome and Safari.

Works in Firefox and Chrome mobile views.

### DIFF
--- a/css/css.css
+++ b/css/css.css
@@ -118,7 +118,7 @@ code {
   color: #c7254e;
   background-color: #f9f2f4;
   border-radius: 4px;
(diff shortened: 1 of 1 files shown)


privatezero:
Looks good to me in Chrome and Firefox on Ubuntu!

ablwr:
good enough for me

PR #291, opened 2017-12-11
consistent aliases used for strip metadata

ablwr:
Strip Metadata was using different versions of the same thing in the line and in the description. This resolves that.


retokromer (approved):
LGTM

PR #293, opened 2017-12-31
add note about quantisation rates and MD5

retokromer:
according to @privatezero’s post in AMIA-L

### DIFF
--- a/index.html
+++ b/index.html
@@ -1662,7 +1662,7 @@
       <dl>
         <dt>ffmpeg</dt><dd>starts the command</dd>
         <dt>-i <i>input_file</i></dt><dd>path, name and extension of the input file</dd>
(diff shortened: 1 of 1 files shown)


privatezero (approved):
This looks good to me!  I wonder if maybe we should create another formula for generating a single audio stream md5 (as opposed to a framemd5 for audio). Then we could point people towards the more simple solution of using `-c copy`.

retokromer:
Indeed, I think another, simpler formula is a good idea! This one is coherent with the idea to have per image frame and per «sound frame» an MD5 value (and the sound values compatible with _BWF MetaEdit_ which, I guess, is frequently used in the wild).

kfrn:
Yeahagahahaha

Issue #298, opened 2018-01-24
Extend recipe: concatenating different file types


kfrn:
We now have a recipe for [concat-ing files of different types](https://amiaopensource.github.io/ffmprovisr/#join_different_files) - however, it only works properly if the input files have the same dimensions (e.g., 720x576).

It would be good to amend the entry to also show how to use the concat filter with inputs of different resolutions, i.e., using a scale filter prior to concatenation.

I can do this, but I thought I'd open this issue in case anyone else is keen! Maybe someone is looking for a contribution to make?

_**Edit**_: it may also be worthwhile to add an example of framerate conversion/normalisation prior to concatenation. 

- [x] Add example of resizing prior to concatenation ([e.g.](https://amiaopensource.github.io/ffmprovisr/#SD_HD))
- [x] Add example of framerate conversion prior to concat (e.g., [`-r` flag](https://ffmpeg.org/ffmpeg.html#Video-Options))

todrobbins:
Does the above recipe work for different codecs besides different file types? Just a newb question...

privatezero:
Sure does!

kfrn:
Yep! The inputs can be different codecs, containers, indeed different in most ways - frame dimensions are a special case. If you like, feel free to adjust the wording to make things clearer.

kfrn:
Okay, I went ahead and changed the wording to hopefully make that a bit clearer!

kfrn:
Explanation about resizing prior to concatenation added in 8f146ecdf6b8388b3f5ebf2f3b5770e487d6aafc  

I'll add framerate conversion next, though as always, someone else should feel free to pick it up if they so desire.

retokromer:
> I'll add framerate conversion next

In my opinion, this would be indeed useful. That way, the full video, composed by different sources, can be encoded with all parts at the natural speed.

kfrn:
Concatenating files of different framerates added in #308.

The one thing I haven't done yet is consider PAR - I'm pretty sure that's a special case too.

retokromer:
I did test your command by adding a padding, because I mixed not only file types, but also image ratios.

kfrn:
Yes, aspect ratio is covered. But I had the idea that mismatches in pixel aspect ratio could cause problems. I thought I ran into this at some stage, but maybe I misremember. I'll see if I can recreate it.

kfrn:
[...]

PR #299, opened 2018-01-25
Add explanation about input files for join_different_files recipe

kfrn:
Per https://github.com/amiaopensource/ffmprovisr/issues/298#issuecomment-360288597, to make it clearer that it's not just the container types that may differ!

att: @todrobbins


privatezero:
Makes sense! Looks good to me!

retokromer (approved):
LGTM

retokromer:
You rock, @kfrn!

PR #306, opened 2018-02-10
unify coding style

retokromer:
- fix missed `[ ... ]`
- unify quoting as done in AMIA Open Source (no distinction between constants and variables)


kfrn:
LGTM, cheers @retokromer !

PR #307, opened 2018-02-10
Concatenate files of different resolutions

kfrn:
As discussed in #298, this extends the **concat different file types** entry to explain how to work with files of different resolutions.

I've tested the new commands, but would appreciate someone else sanity-checking them too!


kfrn:
I've also made a CSS tweak - the `h4`s are now a bit bigger. Previously they didn't really stand out from the paragraph text enough, IMO.

### Before
h4-before

### After
h4-after

privatezero:
Looks great!! I love how you also included info for hd/sd!

index.html
```diff
@@ -949,6 +949,15 @@ <h3>Join files together</h3>
...
+      <p><code>ffmpeg -i input1.avi -i input2.mp4 -filter_complex "[0:v:0] scale=1920:1080:flags=lanczos [rescaledVideo], [rescaledVideo] [0:a:0] [1:v:0] [1:a:0] concat=n=2:v=1:a=1 [video_out] [audio_out]" -map "[video_out]" -map "[audio_out]" <i>output_file</i></code></p>
+      <p>However, this will only have the desired visual output if the inputs have the same aspect ratio. If you wish to concatenate an SD and an HD file, you will also wish to pillarbox the SD file while upscaling. (See the <a href="https://amiaopensource.github.io/ffmprovisr/#SD_HD_2">Convert 4:3 to pillarboxed HD</a> command). The full command would look like this:</p>
+      <code>ffmpeg -i input1.avi -i input2.mp4 -filter_complex "[0:v:0] scale=1440:1080:flags=lanczos, pad=1920:1080:(ow-iw)/2:(oh-ih)/2 [toHDvideo], [toHDvideo] [0:a:0] [1:v:0] [1:a:0] concat=n=2:v=1:a=1 [video_out] [audio_out]" -map "[video_out]" -map "[audio_out]" <i>output_file</i></code>
```
retokromer: Should this line not be inside `<p>` and `</p>` as well?

index.html
```diff
@@ -949,6 +949,15 @@ <h3>Join files together</h3>
...
+      <h4>Variation: concatenating files of different resolutions</h4>
+      <p>To concatenate files of different resolutions, you need to resize the videos to have matching resolutions prior to concatenation. The most basic way to do this is by using a scale filter and giving the dimensions of the file you wish to match:</p>
+      <p><code>-vf scale=1920:1080:flags=lanczos</code></p>
+      <p>(The Lanczos scaling algorithm is recommended, as it is slower but better than the default bilinear algorithm).</p>
+      <p>The rescaling should be applied just before the point where the streams to be used in the output file are listed. Select the stream you want to rescale, apply the filter, and assign that to a variable name (<code>rescaledVideo</code> in the below example). Then you use this variable name in the list of streams to be concatenated.</p>
+      <p><code>ffmpeg -i input1.avi -i input2.mp4 -filter_complex "[0:v:0] scale=1920:1080:flags=lanczos [rescaledVideo], [rescaledVideo] [0:a:0] [1:v:0] [1:a:0] concat=n=2:v=1:a=1 [video_out] [audio_out]" -map "[video_out]" -map "[audio_out]" <i>output_file</i></code></p>
```
retokromer: I suggest to use `rescaled_video` in order to have a consistent naming scheme.

retokromer (commented):
LGTM (sorry, 2 cd/m²)

index.html
kfrn: Nice catch! Fixed in a9c7de6, which also corrects another recipe in the same way.

index.html
kfrn: Fixed in afb7b68. I always forget that we snake-case variable names!

retokromer (approved):
Thank you very much indeed, @kfrn!

ablwr:
yayayayayayyayyayaya

PR #308, opened 2018-02-10
Add info on concatenating files of different framerates

kfrn:
As discussed in #298, this adds more info to the **concat different file types** entry to explain about concating files with different frame rates.

In the recipe, I've suggested the user to use the `fps` and `atempo` filters, slightly different than what is described in the [modify speed](https://amiaopensource.github.io/ffmprovisr/#modify_speed) command. I wasn't getting the desired results with the `setpts` filter, but I admit that that could be partly a late-night-tired mixup.  
However, I'm not sure if `atempo` is the best pick to use with `fps` - @privatezero and others, what do you think?  
For reencoding the audio, I followed the **modify speed** recipe by using a multiplier consisting of the output speed as the numerator and the input speed as the denominator. 

In ffmpeg there seem to be several ways in which to change speed, but I haven't got my head around all of the pros and cons. These are the speed/fps modifiers that I know about:
- [`fps`](https://ffmpeg.org/ffmpeg-filters.html#fps-1) - converts the video to specified constant frame rate by duplicating or dropping frames as necessary
   - as above: is it correct/idiomatic to pair this with `atempo` to handle the audio?
- [`-r`](https://ffmpeg.org/ffmpeg.html#Video-Options) - when used as an output option, duplicates or drops input frames to achieve constant output frame rate.
   - seems not functionally different from `fps`, but I guess they are different algorithms? Anyone know?
- [`setpts`](https://ffmpeg.org/ffmpeg-filters.html#setpts_002c-asetpts) - "changes the PTS (presentation timestamp) of the input frames"
   - goes hand-in-hand with audio filter `asetpts`. (Should we be using them together in the **modify speed** recipe?)
   - if a big difference in fps is used, this results in slow or fast motion. For me the wording of the docs is a little confusing - to me it implies that no frames are dropped/duplicated, simply the timing of them is changed. But I ran a few tests and it seems like the original framerate is maintained, with frames dropped/duplicated as necessary.

Now I'm wondering if there's a way to change speed without dropping/repeating frames - where all the frames are retained, but just the frame rate is affected. In other words, change the stream metadata rather than re-encode.

Anyone got thoughts about this stuff???

_(Also, not sure why the previous lot of commits are showing up, even though their diff isn't part of this PR ... sorry ...)._


index.html
```diff
@@ -957,7 +957,13 @@ <h4>Variation: concatenating files of different resolutions</h4>
...
-      <p>Here, the first input an SD file which needs to be upscaled to match the second input, which is 1920x1080. The scale filter enlarges the SD input to the height of the HD frame, keeping the 4:3 aspect ratio; then, the video is pillarboxed within a 1920x1080 frame.</p>
+      <p>Here, the first input is an SD file which needs to be upscaled to match the second input, which is 1920x1080. The scale filter enlarges the SD input to the height of the HD frame, keeping the 4:3 aspect ratio; then, the video is pillarboxed within a 1920x1080 frame.</p>
+      <h4>Variation: concatenating files of different framerates</h4>
+      <p>If the input files have different framerates, then the output file may be of variable framerate. To explicitly attain an output file of constant framerate, you may wish convert an input (or multiple inputs) to a different framerate prior to concatenation.</p>
```
retokromer: > explicitly attain an output

obtain

retokromer:
**Video:** I guess, it depends on the container/codec and on the player. I guess, some do allow variable frame rates (e.g. WebM/VP8) and others not.

**Audio:** We usually work with the Rubberband audio filter, because this gives us the best results. However, it needs to have FFmpeg installed with Rubberband:
```
brew install ffmpeg --with-rubberband
```
This could be the topic of an _ad hoc_ recipe.

index.html
```diff
@@ -957,7 +957,13 @@ <h4>Variation: concatenating files of different resolutions</h4>
...
+      <p>You can speed up or slow down a file using the <code>fps</code> and <code>atempo</code> filters (see also the <a href="https://amiaopensource.github.io/ffmprovisr/#modify_speed">Modify speed</a> command).</p>
+      <p>Here's an example of the full command, in which input1 is 30fps, input2 is 25fps, and 25fps is the desired output speed.</p>
+      <p><code>ffmpeg -i input1.avi -i input2.mp4 -filter_complex "[0:v:0] fps=fps=25 [video_to_25fps]; [0:a:0] atempo=(25/30) [audio_to_25fps]; [video_to_25fps] [audio_to_25fps] [1:v:0] [1:a:0] concat=n=2:v=1:a=1 [video_out] [audio_out]" -map "[video_out]" -map "[audio_out]" <i>output_file</i></code></p>
```
retokromer: I’m sorry, I would suggest `input_1`, etc.

retokromer: Question: if you concatenate `.avi` and `.mp4`, do you not need a parameter which enforces the higher quality? I guess, otherwise the lower is chosen, but I may be wrong.

index.html
kfrn: Good catch!

index.html
kfrn: Will snake-case the names.   
I'm not totally clear on what you are asking @retokromer! Are you able to rephrase?

retokromer: @kfrn If one mixes different containers/codecs, without forcing the quality (e.g. with `-pix_fmt` for the image), then FFmpeg try to chose the best possible, but I have the **_feeling_** that sometimes it fails. I had a discussion about that some years ago with Nicolas George in Paris. I **_guess_** it’s more secure to impose. You can impose the higher quality to the lower and it might make the lower worst. You can impose the lower quality to the higher and you always will loose quality.

(I don’t have the time to delve into this in a scientific way now. Possibly you have already? I suggest to commit this PR. We can always change later, when needed.)

kfrn: I think I'm with you now ... so, unless you explicitly specify otherwise, it's the output file type which will determine the file quality. e.g., if you have `output.mp4` the video will be encoded as H.264 with the default settings ([i.e.](https://trac.ffmpeg.org/wiki/Encode/H.264), crf 23, preset medium).  

If you mean `-pix_fmt` specifically, I actually tested what happens if the two inputs have different chroma subsampling schemes - but I can't remember the result. :-/  I can probably test it tonight.

ablwr:
Wow @kfrn!

I don't have any insight/opinions here, but looks like good work to me.

index.html
kfrn: **Update**: I didn't test things extensively, but when concat-ing a 4:2:2 and a 4:2:0 file, I'm observing 4:2:2 outputs.

retokromer (approved):
LGTM

index.html
retokromer: Yes, I guess we can merge it. I’ll test a little with WebM/VP8 later this week, as we have a client that wishes that combination and the amateur part of the collection contains different frame rates.

kfrn: Cool, please report back! :-)

kfrn:
Code fixes made!   
However, I'd be really happy if someone else can test before merge.  
Also very interested if anyone has an opinion on the ffmpeg-native (i.e., not from `rubberband`) audio filters.

retokromer (approved):

retokromer:
The quality of the native audio filters are fully OK for **access files**. We use `rubberband` also for those, but it’s mainly to keep our workflow simple. We do prefer `rubberband` for **mezzanine files**, because one can do more, and it gives often a better result as well. Often, not always: it’s a wee bit similar to Lanczos, which usually gives a better result, but there are counterexamples…

I have tested @kfrn‘s command before approving it.

kfrn:
I had forgotten that this was still open, so cheers, @retokromer! If you've tested and approved then we can merge, I think.

PR #318, opened 2018-04-15
removes calls to google fonts CDN

ablwr:
Up for discussion, but I find that the short lag it takes to load and render fonts from the Google CDN when running ffmprovisr locally to be really annoying. Removing it will make the website marginally faster and look identical offline and online, and not throw errors when offline.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3,7 +3,6 @@
 <head>
   <title>ffmprovisr</title>
   <meta name="viewport" charset="utf-8" content="text/html, width=device-width, initial-scale=1">
-  <!-- <link href="https://fonts.googleapis.com/css?family=Montserrat%7CMerriweather" rel="stylesheet" type="text/css"> -->
   <link rel="stylesheet" href="css/css.css">
   <link rel="icon" href="img/vhs.ico">
   <script src="js/jquery.min.js"></script>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -3,7 +3,7 @@
 <head>
   <title>ffmprovisr</title>
   <meta name="viewport" charset="utf-8" content="text/html, width=device-width, initial-scale=1">
-  <link href="https://fonts.googleapis.com/css?family=Montserrat%7CMerriweather" rel="stylesheet" type="text/css">
+  <!-- <link href="https://fonts.googleapis.com/css?family=Montserrat%7CMerriweather" rel="stylesheet" type="text/css"> -->
```
retokromer: I would suggest to delete, instead of commenting out.

retokromer:
Is `font-family: sans-serif;` in `body` not sufficient?

ablwr:
Ah, mistake on my part! Didn't go back in and check the index.html code.

And point taken! I will clean up the code a bit. Unsure if people are interested in more typeface diversity. This keeps the code blocks in monospace type.

ablwr:
@retokromer feelings?

retokromer (approved):
By reading the code seems fine to me.

Issue #211, opened 2017-07-20
Wishlist: search functionality


bturkus:
Hi,
I was wondering about the possibility of adding a search function to ffmprovisr. Now that there are so many commands on there, I've actually had some difficulty locating specific items. If I could search for, say "yuv420p," or whatever else, I think that would be really helpful.
Thanks,
Ben

ablwr:
Ohhh... hmmm... I think this might be possible, yeah. For AVAA I use this client-side javascript library (which means it can still be hosted simply on Github pages and not require a server). Maybe we can integrate that. Anything bigger would require a bigger hosting platform.

retokromer:
As one HTML page contains all the information, it should not be difficult to program a solution that works both online and offline.

ablwr:
ctrl-f

retokromer:
Hmm… I would prefer something more user friendly.

kfrn:
Recently, I was thinking about the idea of being able to tag different commands. But this would require quite a restructuring of ffmprovisr (as, probably, would search ability).

retokromer:
I agree. The Cable Bible could be restructured as well.

ablwr:
Joking about ctrl-f; it doesn't work.

I support searching and tagging! I agree that tagging would require some restructuring -- do you have something in mind? Search probably wouldn't require a restructuring, but could (if it is overall better for the site). 

Ideas?

Or: Ideas for how we can best brainstorm up a solution together?

retokromer:
> Joking about ctrl-f; it doesn't work.

It cannot work ;-)

> Ideas?

I guess, I would split it into the single recipes. As we give the possibility to install and update locally via Homebrew/Linuxbrew, that should be fine also when the internet connection is bad.

ablwr:
Like replicating the way [AVAA](https://bavc.github.io/avaa/) is structured?

retokromer:
I should check that more carefully, but I have the impression that this is completely hard-coded. I would suggest a solution that creates a query’s answer on the fly. (Because it’s easier to maintain.)

ablwr:
AVAA uses lunr.js, a lightweight client-side search service built in JavaScript!

retokromer:
Thank you for the information! Yes, something like that. (My websites use a Perl script and a MariaDB database to generate most of the pages on the fly since more than 20 years.)

ablwr:
[...]

PR #319, opened 2018-04-15
adds open all button

ablwr:
Long-standing quick hack for the searchability feature request from @bturkus. This button allows for user to open all recipes, so the page can then be searched using browser-based ctrl-f.

### DIFF
--- a/index.html
+++ b/index.html
@@ -19,6 +19,7 @@
   <nav class="sidebar well">
     <h2 class="heading">Table of Contents</h2>
     <a href="#about"><div class="contents-list">About this resource</div></a>
+    <div id="open-all" class="contents-list">Open all recipes</div>
     <a href="#basics"><div class="contents-list">FFmpeg basics</div></a>
     <a href="#concepts"><div class="contents-list">Advanced FFmpeg concepts</div></a>
     <a href="#rewrap"><div class="contents-list">Change container (rewrap)</div></a>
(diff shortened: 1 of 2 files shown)


retokromer:
Somewhere we should also give explicitly the CTRL-F hint to make https://github.com/amiaopensource/ffmprovisr/issues/211 possible, I guess.

ablwr:
Oh yes -- seeking feedback on placement of this button and if it should be somewhere else, like on the main section after the info?

retokromer:
I don’t have a strong opinion on the placement.

If I click a second time, nothing new happens. Would it be useful to have an open/close button, instead of a open-only button?

ablwr:
Yes! I will make it into a toggle. Thank you! But does this generally sound
like a good idea?

retokromer:
>But does this generally sound like a good idea?

Yes, it does!

privatezero:
Yah, I definitely think this sounds like a good plan!

ablwr:
OK y'all sorry for being so lazy, I did it right this time!

index.html
```diff
@@ -19,6 +19,7 @@ <h1>➺ ffmprovisr ❥</h1>
   <nav class="sidebar well">
     <h2 class="heading">Table of Contents</h2>
     <a href="#about"><div class="contents-list">About this resource</div></a>
+    <div id="open-all" class="contents-list">Open all recipes</div>
```
retokromer: I would say `Open/close all recipes`.

JonnyTech: `show / hide`?

ablwr: updated with show/hide!

retokromer (approved):

bturkus:
hey I totally spaced on this, but thanks so much! I'm using it already!

PR #326, opened 2018-06-09
uniform external links

retokromer:


privatezero (approved):
looks good!

ablwr:
Whoops! Notoriously bad at remembering that we do this. Thanks @retokromer 💃

PR #336, opened 2018-07-28
unify style

retokromer:
I suggest to put both `lavfi` filters into quotation marks – or use both without.
(Context: It has confused a student.)


kieranjol (approved):
makes sense, LGTM!

PR #350, opened 2018-11-07
rm "optionally" for Matroska-specific description

ablwr:
This recipe explicitly calls for making a Matroska file, so I don't think it should have the part about optionally choosing something else, which makes sense for other recipes.


kieranjol (approved):

retokromer (approved):

kfrn:
Good catch! Thanks all!

PR #353, opened 2018-11-09
Add get checksum on streams

ablwr:
As per the ever-helpful FFmpeg twitter (https://twitter.com/FFmpeg/status/1060662187448950784), I added their recommended recipe. This isn't even, in practice, that dissimilar to the above script for grabbing the md5 of a video stream, but I think it's different enough to merit contribution. Also fixed a typo.


retokromer (approved):
LGTM

PR #361, opened 2018-12-21
update info for Windows

retokromer:
Today, it’s usually easier to use the apps rather than the subsystem.


privatezero (approved):
lgtm!

kfrn:
Right, I'd forgotten that you can install the subsystem via app now, good catch!

retokromer:
It’s much easier.

PR #364, opened 2019-01-17
uniform URLs

retokromer:
- use `https://` when possible
- add missing `target="_blank"` to external links
- uniform internal links


kfrn:
Looks good! Great idea to just use the IDs for self-links.

retokromer:
Thank you, @kfrn! The majority were already IDs for self-links, yet one was a full URL, causing an error when used locally as offline app.

ablwr (approved):

PR #367, opened 2019-02-01
Silence trim

privatezero:
Adds recipes for trimming silence from front/end of audio file

### DIFF
--- a/index.html
+++ b/index.html
@@ -1219,12 +1219,13 @@
     <!-- ends Excerpt from end -->
 
     <!-- Trim start silence -->
-      <label class="recipe" for="trim_start_silence">Trim silence from beginning of an audio file</label>
-      <input type="checkbox" id="trim_start_silence">
-      <div class="hiding">
-      <h3>Remove silent portions at the beginning of an audio file</h3>
-      <p><code>ffmpeg -i <em>input_file</em>-af silenceremove=start_threshold=-57dB:start_duration=1:start_periods=1<em>output_file</em></code></p>
-      <p>This command will automatically remove silence at the beginning of an audio file. The threshold for what qualifies as silence can be changed - this example uses anything under -57dB, which is a decent level for accounting for analogue hiss.</p>
+    <label class="recipe" for="trim_start_silence">Trim silence from beginning of an audio file</label>
+    <input type="checkbox" id="trim_start_silence">
+    <div class="hiding">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1218,6 +1218,52 @@ <h3>Excerpt from end</h3>
...
     <!-- ends Excerpt from end -->
 
+    <!-- Trim start silence -->
+    <label class="recipe" for="trim_start_silence">Trim silence from beginning of an audio file</label>
+    <input type="checkbox" id="trim_start_silence">
+    <div class="hiding">
+      <h3>Remove silent portion at the beginning of an audio file</h3>
+      <p><code>ffmpeg -i <em>input_file</em> -af silenceremove=start_threshold=-57dB:start_duration=1:start_periods=1  -c:a <em>your_codec_choice</em> -ar <em>your_sample_rate_choice</em> <em>output_file</em></code></p>
+      <p>This command will automatically remove silence at the beginning of an audio file. The threshold for what qualifies as silence can be changed - this example uses anything under -57 dB, which is a decent level for accounting for analogue hiss.</p>
+      <p>Note: Since this is applying a filter, you can not use codec copy for your output. If you do not specify an output sample rate and codec this command will use <a href='#codec-defaults'>the defaults for your output format</a>. Take care that you are getting your intended results!</p>
```
kfrn: Nitpick, but it might be clearer to say something like "Since this recipe involves a filter, the audio stream will be re-encoded". Your call though

kfrn (approved):
Awesome! I haven't tested the actual recipes, but everything looks good to me.

retokromer (approved):

index.html
privatezero: I dig

retokromer (approved):
I’m not English speaker, yet LGTM.

privatezero:
Hmm, one other question for the group - it isn't strictly necessary for the user to specify `-ar` as FFmpeg should pull that from the source, even with the transcode. I put that in as I can see one of the big use cases for this is creating lower sample rate mezzanine files, but maybe it would be cleaner without??

retokromer:
Good point! I guess it would be clearer without it in the main recipe, but I suggest to add this information as a comment.

privatezero:
Aight! removed `-ar` from recipe. Hopefully changed wording makes sense

retokromer (approved):
LGTM

PR #372, opened 2019-02-18
complete information

retokromer:
- the range [0,51] is valid for 8-bit per channel content
- the range [0,63] is valid for 10-bit per channel content

(I suggest to not add the later, because IRL it’s a little absurd to choose 10-bit per channel and 4:2:0 subsampling. In my opinion, it would just be a useless complication for the end user to introduce another pixel format here just for that.)


privatezero (approved):
looks good! I agree about not adding the extra complication.

kieranjol:
LGTM

kfrn (approved):
Nice one!

ablwr (approved):

PR #373, opened 2019-02-28
unify syntax of recipe with syntax of the index

retokromer:
- `reencoding` -> `re-encoding`
- and delete an unneeded space before a tag


privatezero (approved):
lgtm! thanks!

PR #387, opened 2019-07-17
Adds mapping all audio to access script

ablwr:
Connected to https://github.com/amiaopensource/ffmprovisr/issues/383

@bturkus I can't request a review from you (???) but I'd like one.

PR #389, opened 2019-07-19
Revert "Adds mapping all audio to access script"

ablwr:
Reverts amiaopensource/ffmprovisr#387.

I have been testing this script and I don't think it's quite good enough to be recommended. The `map` feature is too broad and it can accidentally create files with only audio. This works for some use cases but not broadly enough to be a good solution for all users.

It'd be nice to spend more time sorting this out but I have other responsibilities.

### DIFF
--- a/index.html
+++ b/index.html
@@ -294,7 +294,7 @@
     <input type="checkbox" id="transcode_h264">
     <div class="hiding">
       <h3>Transcode to H.264</h3>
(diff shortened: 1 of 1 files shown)


ablwr:
@retokromer this affects your subsequent PR and release

retokromer (approved):

retokromer:
@ablwr No problem! I will make a new release as soon as the two PR are merged.

ablwr:
@privatezero notes that it should be okay if `-map 0:v` is added, and I tested that and it seems to be true. I think this needs testing more broadly to be worth adding in, though, so I am going to merge the revert.

Issue #392, opened 2019-07-30
MKV -> mp4 rewrap command possibly misleading?


privatezero:
I've just gotten some user feedback that the recipe at https://amiaopensource.github.io/ffmprovisr/index.html#mkv_to_mp4 is a little problematic from a user perspective in our expected use case of libraries/archives.

Having the `-c:v copy` flag in the command will cause this to fail in the event of someone trying to run it on an ffv1/mkv file. This _seems_ to me at least to be a problem as within the target audience of ffmprovisr, probably a lot of people will be trying to find commands to use on that flavor of file.

Could this possibly be deleted as it is a little redundant with the basic rewrap command and the access mp4 command? Or clarified/cross linked more fully?

kieranjol:
I'm in favour of deletion. I was initally going to suggest switching to something more flexible like mov, but I think that the basic_rewrap recipe makes this pretty redundant, and I agree on the confusion. Thanks for relaying the user feedback!

kfrn:
Yep, I think the functionality is covered by the [basic rewrap](https://amiaopensource.github.io/ffmprovisr/index.html#basic-rewrap) command - I vote to remove [MKV to MP4](https://amiaopensource.github.io/ffmprovisr/index.html#mkv_to_mp4).

Happy for one of you to go ahead, otherwise I can do this when I'm at home tonight

privatezero:
Cool! I agree with ya both! Thanks @kfrn! I'd do it, but I'm wrapping up my current job this week so am swamped with transition prep!

ablwr:
In full support!

retokromer:
Nice catch!

retokromer:
Is this issue still one?

privatezero:
I think this can be closed as the related formula was deleted!

PR #393, opened 2019-07-31
Remove MKV → MP4 command

kfrn:
Addressing user feedback received by Andrew Weaver (@privatezero).

When the source file is ffv1 in MKV, the command will fail on the `-c:v copy` flag, since MP4s can't house ffv1.
Therefore, the command is misleading, and has limited value from an archival perspective.

We already have a basic rewrap command (which includes a note addressing the possibility of stream/wrapper incompability), so I think our bases are covered with that.

See https://github.com/amiaopensource/ffmprovisr/issues/392

### DIFF
--- a/index.html
+++ b/index.html
@@ -213,28 +213,6 @@
     </div>
     <!-- End Basic rewrap command -->
 
-    <!-- MKV to MP4 -->
-    <label class="recipe" for="mkv_to_mp4">Convert Matroska (MKV) to MP4</label>
-    <input type="checkbox" id="mkv_to_mp4">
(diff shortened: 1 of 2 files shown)


kieranjol (approved):

retokromer (approved):

ablwr (approved):

retokromer:
@kfrn I added a commit to update also the list. So we don’t forget it.

privatezero (approved):

kfrn:
Thanks @retokromer, I forgot entirely!

PR #394, opened 2019-08-02
Add note to Stream Mapping entry about mapping with a failsafe

kfrn:
Following Ben Turkus' comment here:
https://github.com/amiaopensource/ffmprovisr/issues/383#issuecomment-516847829

**Apologies!** I accidentally pushed this to master (that is, the `gh-pages` branch - now reverted). Maybe we should turn on branch protection, heh.

### DIFF
--- a/index.html
+++ b/index.html
@@ -178,8 +178,12 @@
         <li><code>-map 0:0 -map 0:2</code> means ‘take the first and third streams from the first input file’.</li>
         <li><code>-map 0:1 -map 1:0</code> means ‘take the second stream from the first input file and the first stream from the second input file’.</li>
       </ul>
-      <p>To map <em>all</em> streams in the input file to the output file, use <code>-map 0</code>. However, note that not all container formats can include all stream types: for example, .mp4 cannot contain timecode.</p>
       <p>When no mapping is specified in an ffmpeg command, the default for video files is to take just one video and one audio stream for the output: other stream types, such as timecode or subtitles, will not be copied to the output file by default. If multiple video or audio streams are present, the best quality one is automatically selected by FFmpeg.</p>
+      <p>To map <em>all</em> streams in the input file to the output file, use <code>-map 0</code>. However, note that not all container formats can include all stream types: for example, .mp4 cannot contain timecode.</p>
+      <h4>Mapping with a failsafe</h4>
+      <p>To safely process files that may or may not contain given a type of stream, you can add a trailing <code>?</code> to your map commands: for example, <code>-map 0:a?</code> instead of <code>-map 0:a</code>.</p>
+      <p>This makes the map optional: audio streams will be mapped over if they are present in the file—but if the file contains no audio streams, the transcode will precede as usual, minus the audio stream mapping. Without adding the trailing <code>?</code>, FFmpeg will exit with an error on that file.</p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -178,8 +178,13 @@ <h3>Stream mapping</h3>
...
       <p>When no mapping is specified in an ffmpeg command, the default for video files is to take just one video and one audio stream for the output: other stream types, such as timecode or subtitles, will not be copied to the output file by default. If multiple video or audio streams are present, the best quality one is automatically selected by FFmpeg.</p>
+      <p>To map <em>all</em> streams in the input file to the output file, use <code>-map 0</code>. However, note that not all container formats can include all stream types: for example, .mp4 cannot contain timecode.</p>
+
+      <h4>Mapping with a failsafe</h4>
+      <p>To safely process files that may or may not contain given a type of stream, you can add a trailing <code>?</code> to your map commands: for example, <code>-map 0:a?</code> instead of <code>-map 0:a</code>.</p>
+      <p>This means that audio streams will be mapped over if they are present in the file—but if the file contains no audio streams, the transcode will precede as usual, minus the audio stream mapping. Without adding the trailing <code>?</code>, FFmpeg will skip over files which don't contain an audio track.</p>
```
kfrn: It skips the files rather than failing on them or throwing an error, right @bturkus? I haven't actually tried it myself. The [ffmpeg docs](https://ffmpeg.org/ffmpeg.html#Advanced-options) suggest that it fails?

Also, let me know if I've expressed this well!

index.html
```diff
@@ -178,8 +178,13 @@ <h3>Stream mapping</h3>
         <li><code>-map 0:0 -map 0:2</code> means ‘take the first and third streams from the first input file’.</li>
         <li><code>-map 0:1 -map 1:0</code> means ‘take the second stream from the first input file and the first stream from the second input file’.</li>
       </ul>
-      <p>To map <em>all</em> streams in the input file to the output file, use <code>-map 0</code>. However, note that not all container formats can include all stream types: for example, .mp4 cannot contain timecode.</p>
       <p>When no mapping is specified in an ffmpeg command, the default for video files is to take just one video and one audio stream for the output: other stream types, such as timecode or subtitles, will not be copied to the output file by default. If multiple video or audio streams are present, the best quality one is automatically selected by FFmpeg.</p>
+      <p>To map <em>all</em> streams in the input file to the output file, use <code>-map 0</code>. However, note that not all container formats can include all stream types: for example, .mp4 cannot contain timecode.</p>
+
```
retokromer: I guess, the empty line inside the recipe is not needed.

retokromer (approved):
1 cd/m² otherwise LGTM.

index.html
kfrn: Now fixed in 10341e1 - cheers :-)

index.html
kieranjol: I think explains it well. Maybe specify ffmpeg will exit with an error and skip over the file?

bturkus: This is really really great; thanks @kfrn!!

kfrn: Thanks chaps! I've just amended the wording, would you take another look? 🙏 Cheers!

kieranjol (approved):

index.html
kieranjol: It looks good to me!

ablwr:
I turned on branch protection!

PR #396, opened 2019-09-26
simplify recipe extraction

### LINKED ISSUE #377: modify the use of the `<h3>` tag
I guess, it’s not possible to satisfy https://github.com/amiaopensource/ffmprovisr/issues/374 with a one-liner, but tedious looping is needed, because the use of `<label class="recipe" for="...">` and `<h3>` are not univoque.

- A dirty hack could be to add a pseudo-class to the main `<h3>` of the recipes.
- In my opinion, a better solution would be to use the `<h3>` tag (or another heading tag) exclusively for the main title of the recipes.

In both cases the `index.html` file could then be parsed by something like:

```sh
cat index.html \
  | grep -E '<h3>.*</h3>|<p><code>.*</code></p>' \
  | sed 's/.*<h3>\(.*\)<\/h3>/# \1/' \
  | sed 's/.*<code>\(.*\)<\/code>/\1/' \
  | grep -v '\*\*\*' \
  | sed -e 's/<[^>]*>//g'
```

The command is composed by the following piped elements:

- read the `index.html` file content
- select the relevant lines (title and command)
- replace the spaces before the `<h3>` tag by the comment sign `# `
- remove the spaces before the `<code>` tag
- remove the sample example
- remove the tags inside the commands, such as `<em>`

retokromer:
should resolve https://github.com/amiaopensource/ffmprovisr/issues/377

### DIFF
--- a/css/css.css
+++ b/css/css.css
@@ -105,7 +105,7 @@ h2 {
   margin: 6px 0px 12px 0px;
 }
 
-h3 {
+h3, h5 {
   font-size: 1.5em;
 }
 
--- a/index.html
+++ b/index.html
@@ -72,7 +72,7 @@
       <label class="recipe" for="basic-structure">Basic structure of an FFmpeg command</label>
       <input type="checkbox" id="basic-structure">
       <div class="hiding">
-        <h3>Basic structure of an FFmpeg command</h3>
+        <h5>Basic structure of an FFmpeg command</h5>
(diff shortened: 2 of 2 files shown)


kfrn (approved):
I'm not in love with this from a semantic HTML point-of-view - but it'll get the job done for the minute :)

ablwr:
Yeah a lot could be improved upon :grimacing: but a larger issue than this one

kfrn:
Yeah, maybe one rainy day eh ...

retokromer:
>I'm not in love with this from a semantic HTML point-of-view

Me neither, that’s actually the reason why it was waiting! In my opinion, the whole document could gain from a semantic review.

ablwr:
It'd be nice if someone could file an issue making that request!

retokromer:
I was working on a PR, but had to slow down work during the Summer for health reasons.

Issue #397, opened 2019-09-26
Semantic HTML


kfrn:
It's come up a couple of times, most recently [here](https://github.com/amiaopensource/ffmprovisr/pull/396) - when time allows, it'd be great to do a pass for semantic HTML.

retokromer:
Is someone working on this?

ablwr:
It's open for someone to work on this

PR #398, opened 2019-10-03
Changes h3 to h5

Anushka-codergirl:
Work done!

### DIFF
--- a/index.html
+++ b/index.html
@@ -243,7 +243,7 @@
     <label class="recipe" for="to_prores">Transcode to deinterlaced Apple ProRes LT</label>
     <input type="checkbox" id="to_prores">
     <div class="hiding">
-      <h3>Transcode into a deinterlaced Apple ProRes LT</h3>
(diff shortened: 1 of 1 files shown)


ablwr:
Hi @Anushka-codergirl! Thanks for making a PR :)

@retokromer would this help your script?

I do think it makes sense for headings that are initially hidden to be a lower-level heading. In general, I think the whole page could use a semantic review, as @kfrn suggests in #397.

But I am in support of incremental changes in the right direction!

ablwr (approved):

kfrn:
Thanks for this, Anushka! However, we're looking to make our HTML more semantic, so I don't think this quite works, because now the heading levels are out of step - `h1`, `h2`, then `h5`. So I'd probably keep these as `h3`s.

kfrn:
Oops, I hadn't refreshed the page and seen @ablwr's comment or the merge. Nevermind, then!

retokromer:
@ablwr Not really, I guess, but I’ll update it.

PR #406, opened 2019-11-04
Added "streamhash" recipe.

pjotrek-b:
See subject :grin:

According to new "streamhash" functionality in FFmpeg:
https://www.ffmpeg.org/ffmpeg-formats.html#streamhash-1

### DIFF
--- a/index.html
+++ b/index.html
@@ -1997,6 +1997,47 @@
     </div>
     <!-- ends Get checksum for video/audio stream -->
 
+    <!-- Get checksum for all video/audio streams -->
+    <label class="recipe" for="get_streamhash">Get individual checksums for all video/audio streams ("Streamhash")</label>
+    <input type="checkbox" id="get_streamhash">
+    <div class="hiding">
+      <h5>Get individual checksums for all video/audio streams ("Streamhash")</h5>
+      <p><code>ffmpeg -i <em>input_file</em> -map 0 -f streamhash -hash md5 - -v quiet</code></p>
+      <p>
+      The outcome is very similar to that of "-f hash", except you get one hash
+      per-stream, instead of one (summary) hash.  Another benefit is that you
+      don't have to know which streams, or how many to expect in the source
+      file. This is very handy for hashing mixed born-digital material.
+      </p>
+      <p>
+      This script will perform a fixity check on all audio and video streams in
+      the file and return one hashcode for each one.  This is useful for e.g.
+      be able to change to container/codec format later on and validate it
+      matches the original source.
+      </p>
+      <p>
+      The output is formatted for easily processing it further in any kind of
+      programming/scripting language.
+      </p>
+      <dl>
+        <dt>ffmpeg</dt><dd>starts the command</dd>
+        <dt>-i <em>input_file</em></dt><dd>path, name and extension of the input file</dd>
+        <dt>-map 0</dt><dd>map ALL streams from input file to output. If you omit this, ffmpeg chooses only the first "best" (*) stream: 1 for audio, 1 for video (not all streams).</dd>
+        <dt>-f streamhash -hash md5</dt><dd>produce a checksum hash per-stream, and set the hash algorithm to md5. See the official <a href="https://www.ffmpeg.org/ffmpeg-formats.html#streamhash-1" target="_blank">documentation on streamhash</a> for other algorithms and more details.</dd>
+        <dt>-</dt><dd>FFmpeg syntax requires a specified output, and <code>-</code> is just a place holder. No file is actually created. Choose an output filename to write the hashcode lines into a textfile.</dd>
+        <dt>-v quiet</dt><dd>(Optional) Disables FFmpeg's processing output. With this option it's easier to see the text output of the hashes.</dd>
+      </dl>
+      <p>
+      The output looks like this, for example (1 video, 2 audio streams):
+      <code>
+      0,v,MD5=89bed8031048d985b48550b6b4cb171c
+      0,a,MD5=36daadb543b63610f63f9dcff11680fb
+      1,a,MD5=f21269116a847f887710cfc67ecc3e6e
+      </code>
+      <p class="link"></p>
+    </div>
+    <!-- ends Get checksum for all video/audio streams -->
+
     <!-- QCTools Report -->
     <label class="recipe" for="qctools">QCTools report (with audio)</label>
     <input type="checkbox" id="qctools">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1997,6 +1997,47 @@ <h5>Get checksum for video/audio stream</h5>
...
     <!-- ends Get checksum for video/audio stream -->
 
+    <!-- Get checksum for all video/audio streams -->
+    <label class="recipe" for="get_streamhash">Get individual checksums for all video/audio streams ("Streamhash")</label>
+    <input type="checkbox" id="get_streamhash">
+    <div class="hiding">
+      <h5>Get individual checksums for all video/audio streams ("Streamhash")</h5>
+      <p><code>ffmpeg -i <em>input_file</em> -map 0 -f streamhash -hash md5 - -v quiet</code></p>
+      <p>
+      The outcome is very similar to that of "-f hash", except you get one hash
```
kfrn: Nitpick - I would prefer if the text inside the `<p>` tags was indented inwards (also below)

index.html
```diff
@@ -1997,6 +1997,47 @@ <h5>Get checksum for video/audio stream</h5>
...
+      </p>
+      <dl>
+        <dt>ffmpeg</dt><dd>starts the command</dd>
+        <dt>-i <em>input_file</em></dt><dd>path, name and extension of the input file</dd>
+        <dt>-map 0</dt><dd>map ALL streams from input file to output. If you omit this, ffmpeg chooses only the first "best" (*) stream: 1 for audio, 1 for video (not all streams).</dd>
+        <dt>-f streamhash -hash md5</dt><dd>produce a checksum hash per-stream, and set the hash algorithm to md5. See the official <a href="https://www.ffmpeg.org/ffmpeg-formats.html#streamhash-1" target="_blank">documentation on streamhash</a> for other algorithms and more details.</dd>
+        <dt>-</dt><dd>FFmpeg syntax requires a specified output, and <code>-</code> is just a place holder. No file is actually created. Choose an output filename to write the hashcode lines into a textfile.</dd>
+        <dt>-v quiet</dt><dd>(Optional) Disables FFmpeg's processing output. With this option it's easier to see the text output of the hashes.</dd>
+      </dl>
+      <p>
```
kfrn: Is this one missing a closing tag?

kfrn (changes requested):
Recipe looks great! I'll test it when I'm at home this evening.

Would you mind doing a minor HTML tidy? 😀👍

privatezero:
In the recipe explanation I might add an explicit note that the audio stream is converted to 16 bit uncompressed audio for hashing. Since so much archival audio contains 24 bit uncompressed audio I have seen many people get confused by audio hashes that they expected to match-up because they assumed FFmpeg was keeping input bit-depth.

privatezero:
Just tested recipe and works great! Thanks! Such a useful new filter!

index.html
pjotrek-b: Sure!
You mean `<p>...</p>` instead of:
```
<p>
...
</p>
```

Will do that!

retokromer:
>Such a useful new filter!

Indeed!

index.html
kfrn: In fact, I meant,
```html
<p>
  ....
  ....
</p>
```

retokromer: In my opinion, should be `<p>. . .</p>` without newlines inside the `. . .`.

index.html
pjotrek-b: Indeed! Sorry.
Will run "[HTML tidy](http://www.html-tidy.org/)".

pjotrek-b:
Cleaned up the `<p>`'s :smile:

btw: "tidy" found another missing `</p>` in line 1870 (PowerShell recipe), but I'll fix that after this PR was merged in order not to mess up the commit log.

retokromer (approved):

pjotrek-b:
@privatezero & @retokromer 
Happy to hear you find it useful too!

I had it implemented in order to finally deal with mixed born-digital material, and reduce the "guesswork" code hacks necessary for rewrapping/normalizing unknown source streams.
(Thanks to Moritz Barsnick who wrote the patch)

retokromer:
Possibly we should add a note clarifying that this filter works only on HEAD (until the next FFmpeg release).

index.html
```diff
@@ -1997,6 +1997,40 @@ <h5>Get checksum for video/audio stream</h5>
...
+      <h5>Get individual checksums for all video/audio streams ("Streamhash")</h5>
+      <p><code>ffmpeg -i <em>input_file</em> -map 0 -f streamhash -hash md5 - -v quiet</code></p>
+      <p>The outcome is very similar to that of "-f hash", except you get one hash
+      per-stream, instead of one (summary) hash.  Another benefit is that you
+      don't have to know which streams, or how many to expect in the source
+      file. This is very handy for hashing mixed born-digital material.</p>
+      <p>This script will perform a fixity check on all audio and video streams in
+      the file and return one hashcode for each one.  This is useful for e.g.
+      be able to change to container/codec format later on and validate it
+      matches the original source.</p>
```
kieranjol, line 2013: 'so that it matches'

kieranjol (commented):
Great addition!

index.html
kfrn: I feel like @retokromer gets final word on our style guide ;-), so let's go with his suggestion :-)

index.html
kfrn: Huh, the part I would change is `e.g. be able` -> `e.g. being able`

kfrn:
> btw: "tidy" found another missing `</p>` in line 1870 (PowerShell recipe), but I'll fix that after this PR was merged in order not to mess up the commit log.

I'm fastidious like this too! Love it

kfrn (approved):
Thank you!

ablwr (approved):

ablwr:
Thanks @pjotrek-b and everybody!

todrobbins:
Thanks all! @pjotrek-b 🎉

mercuryswitch:
> In the recipe explanation I might add an explicit note that the audio stream is converted to 16 bit uncompressed audio for hashing. Since so much archival audio contains 24 bit uncompressed audio I have seen many people get confused by audio hashes that they expected to match-up because they assumed FFmpeg was keeping input bit-depth.

Perhaps a silly question but why is it converted to 16bit, is it a limitation of hashing?

privatezero:
@mercuryswitch, this is just an assumption of mine, but I imagine they had to pick a flavor of raw audio to use as a default and went with 16 bit because of its ubiquity?

It isn't an actual limitation, as the FFmpeg stream hashing filters do in fact support hashing other codecs/bit depths, it just has to be specified explicitly to override the defaults, so something to be conscious of as a user.

PR #409, opened 2019-11-20
Update index.html

macasaurusrex:
fix for ticket 399: meta tag validation error. w3 doesn't like chartset in the same tag as viewport so I split them out based on what was described in the ticket

### DIFF
--- a/index.html
+++ b/index.html
@@ -2,7 +2,8 @@
 <html lang="en">
 <head>
   <title>ffmprovisr</title>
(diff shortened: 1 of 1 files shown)


ablwr:
Yesssssss, thank you so much @macasaurusrex !

ablwr (approved):

todrobbins:
Thank you @macasaurusrex!

PR #422, opened 2020-05-02
Update: Adding fade/afade to streams

DaleLore:
This is for Issue: https://github.com/amiaopensource/ffmprovisr/issues/386

### DIFF
--- a/index.html
+++ b/index.html
@@ -754,7 +754,6 @@
     <input type="checkbox" id="sync_streams">
     <div class="hiding">
       <h5>Synchronize video and audio streams</h5>
-
       <p><code>ffmpeg -i <em>input_file</em> -itsoffset 0.125 -i <em>input_file</em> -map 1:v -map 0:a -c copy <em>output_file</em></code></p>
       <p>A command to slip the video channel approximate 2 frames (0.125 for a 25fps timeline) to align video and audio drift, if generated during video tape capture for example.</p>
       <dl>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -729,11 +729,32 @@ <h5>Modify image and sound speed</h5>
...
+      <p class="link"></p>
+    </div>
+    <!-- ends Fade both video and audio streams -->
+
     <!-- Synchronize video and audio streams -->
     <label class="recipe" for="sync_streams">Synchronize video and audio streams</label>
     <input type="checkbox" id="sync_streams">
     <div class="hiding">
       <h5>Synchronize video and audio streams</h5>
+
```
ablwr:
```suggestion

```

ablwr: Don't need this space

ablwr (changes requested):

ablwr (approved):

kieranjol:
This is great!
It could be helpful to explain what the `st ` and `d`values mean? Seems to
be start and duration? Also the ffmpeg documentation says that these are
second values. It could be good to clarify that as the opening explanation
refers for first 30 frames and last 44100 samples, sso I initially thought
that s and d might be frame counts. Also it could be good to explain why a
value of 59 is used. Is it a one minute source video? That would be good to
so that folks would know how to scale the recipe to a file of a different
duration.

PR #425, opened 2020-05-13
Remove jQuery

ablwr:
We were using the jQuery library for a few things, but I replaced these things with plain ol' Javascript.

I did some testing here, but it could definitely use more thorough testing to make sure I cleaned it up right! Pushing this now just so I can move on to other tasks today. ;) But will rush back to fix anything that didn't successfully make the transition.


index.html
```diff
@@ -2864,4 +2863,5 @@ <h3>*****Longer title*****</h3>
 </footer>
 </div><!-- ends "grid" -->
 </body>
+<script src="js/js.js"></script>
```
retokromer, line 2865: Are you sure this should not remain inside the `<head>` rather than at the `<html>` level?

ablwr: Oh thanks, let me move it up! Or it will be invalid html.

privatezero (approved):
Just tested - seems to perform as anticipated on my end

js/js.js
```diff
@@ -1,52 +1,96 @@
...
+}
+
+// open recipe window if a hash is found in URL
+if (window.location.hash) {
+  id = window.location.hash.slice(1)
+  moveToRecipe(id)
+}
+
+// add hash URL when recipe is opened
+let recipes = document.querySelectorAll('label[class="recipe"]')
```
kfrn: (minor) `recipes` (and some of the other `let` variables) can be `const`s, no?

kfrn (approved):
I won't have the chance to test until the evening, but LGTM!

js/js.js
ablwr: yes! thanks @kfrn -- clearly was just being lazy here

retokromer (approved):

PR #426, opened 2020-05-24
fade.v2

DaleLore:
The fade in/out was in reference to #386. 
This update is to (hopefully) better clarify the instructions regarding what st and d were (start and duration). Why the use of 59 (it should be the equivalence of a second according to the video.stackexchange forum). 

Thanks @kieranjol for the recommendations (#422)

### DIFF
--- a/index.html
+++ b/index.html
@@ -738,8 +738,8 @@
       <dl>
           <dt>ffmpeg</dt><dd>starts the command</dd>
           <dt>-i <em>input_file</em></dt><dd>path, name and extension of the input file</dd>
-          <dt>-filter:v "fade=in:st=0:d=1, fade=out:st=59:d=1"</dt><dd>applies a video filter that fades in the first and last 30 video frames</dd>
-          <dt>-filter:a "afade=in:st=0:d=1, afade=out:st=59:d=1"</dt><dd>applies an audio filter that fades in the first and last 30 video frames. <code>st</code> sets the start and <code>d</code> sets the duration.</dd>
+          <dt>-filter:v "fade=in:st=0:d=1, fade=out:st=59:d=1"</dt><dd>applies a video filter that fades in the first and last 30 video frames. <code>st</code> sets the start and <code>d</code> sets the duration. <code>st=59:d=1</code> is equivalent to one second</dd>
+          <dt>-filter:a "afade=in:st=0:d=1, afade=out:st=59:d=1"</dt><dd>applies an audio filter that fades in the first and last 30 video frames and 44100 audio samples. <code>st</code> sets the start and <code>d</code> sets the duration.<code>st=59:d=1</code> is equivalent to one second</dd>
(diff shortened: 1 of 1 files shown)


retokromer (commented):
I would add that the time duration can be expressed either in seconds or by a “timecode”, see https://ffmpeg.org/ffmpeg-utils.html#time-duration-syntax

ablwr:
@bturkus I can't add you as a reviewer but maybe you'd like to review?

index.html
```diff
@@ -738,8 +738,8 @@ <h5>Fade both video and audio streams</h5>
       <dl>
           <dt>ffmpeg</dt><dd>starts the command</dd>
           <dt>-i <em>input_file</em></dt><dd>path, name and extension of the input file</dd>
-          <dt>-filter:v "fade=in:st=0:d=1, fade=out:st=59:d=1"</dt><dd>applies a video filter that fades in the first and last 30 video frames</dd>
-          <dt>-filter:a "afade=in:st=0:d=1, afade=out:st=59:d=1"</dt><dd>applies an audio filter that fades in the first and last 30 video frames. <code>st</code> sets the start and <code>d</code> sets the duration.</dd>
+          <dt>-filter:v "fade=in:st=0:d=1, fade=out:st=59:d=1"</dt><dd>applies a video filter that fades in the first and last 30 video frames. <code>st</code> sets the start and <code>d</code> sets the duration. <code>st=59:d=1</code> is equivalent to one second</dd>
```
kieranjol, line 741: The explanation here assumes that the input is one minute long and 30fps. So it would be good to state that the example is 1 minute long and 30fps, or preferably, remove the reference to 30fps altogether, and instead just refer to seconds. `applies a video filter that fades in the first and last second.` And instead of the last line, perhaps: `<code>st=59:d=1</code> this example uses a start time of 59 as it is a 60 second video`

index.html
```diff
@@ -738,8 +738,8 @@ <h5>Fade both video and audio streams</h5>
...
+          <dt>-filter:a "afade=in:st=0:d=1, afade=out:st=59:d=1"</dt><dd>applies an audio filter that fades in the first and last 30 video frames and 44100 audio samples. <code>st</code> sets the start and <code>d</code> sets the duration.<code>st=59:d=1</code> is equivalent to one second</dd>
```
kieranjol, line 742: Again, I'd remove references to frames here as the code is talking about duration in seconds . I'd also remove the reference to 44100 audio samples, as this would only be accurate if the source video is 44.1kHz, when a lot of video ends up being 48kHz or something else. I'd repeat the clarification about the value `59` referring to this being a 60 second source video.

kieranjol (changes requested):
I'd echo @retokromer suggestions, but i've also left some comments about the ffmpeg code referring to seconds, but the explanation referring to samples/frames.

kfrn:
I'm late to the party here! But I agree with the others that it would be handy to have more generic examples (or more explanation about specific figures used).

kfrn:
Closing because we have another PR, #428.

@DaleLore, in future, would you mind updating the existing PR rather than opening a new one? Thanks!

Issue #424, opened 2020-05-02
Refine definition for fade/afade


ablwr:
@DaleLore or @bturkus,

The fade/afade recipe could use better explanation into what is happening. See @kieranjol's comments [here](https://github.com/amiaopensource/ffmprovisr/pull/422#issuecomment-622791512).

PR #428, opened 2020-05-29
Another attempt at fade in/out documentation

DaleLore:
Hopefully I got all the uses of "frames" out. (Let me know if one snuck in)
And I tried to be more specific by using the example that was used in the stackoverflow forum. 

This is for Issue: OG #386 / First attempt #422 / Now this attempt #424

### DIFF
--- a/index.html
+++ b/index.html
@@ -733,13 +733,13 @@
     <input type="checkbox" id="fade_streams">
     <div class="hiding">
       <h5>Fade both video and audio streams</h5>
-      <p><code>ffmpeg -i <em>input_file</em> -filter:v "fade=in:st=0:d=1, fade=out:st=59:d=1" -filter:a "afade=in:st=0:d=1, afade=out:st=59:d=1" -c:v libx264 -c:a aac <em>output_file</em></code></p>
-      <p>This command fades in and out the first and last second (and 44100 audio samples) on a video.</p>
+      <p><code>ffmpeg -i <em>input_file</em> -filter:v "fade=in:st=IN_POINT:d=DURATION, fade=out:st=OUT_POINT:d=DURATION" -filter:a "afade=in:st=OUT_POINT:d=DURATION, afade=out:st=IN_POINT:d=DURATION" -c:v libx264 -c:a aac <em>output_file</em></code></p>
+      <p>This command fades in and out the first and last second on a video. Change IN_POINT, OUT_POINT, DURATION to the time in seconds (expressed as integers) you want.</p>
       <dl>
           <dt>ffmpeg</dt><dd>starts the command</dd>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -734,12 +734,12 @@ <h5>Modify image and sound speed</h5>
     <div class="hiding">
       <h5>Fade both video and audio streams</h5>
       <p><code>ffmpeg -i <em>input_file</em> -filter:v "fade=in:st=0:d=1, fade=out:st=59:d=1" -filter:a "afade=in:st=0:d=1, afade=out:st=59:d=1" -c:v libx264 -c:a aac <em>output_file</em></code></p>
-      <p>This command fades in and out the first and last 30 video frames and 44100 audio samples on a video</p>
+      <p>This command fades in and out the first and last second (and 44100 audio samples) on a video.</p>
```
kfrn: Wondering where the `44100 audio samples` came from? What if the input isn't 44.1 kHz?

kfrn:
Thanks, @DaleLore!

I wonder if we should make it even more generic?
e.g.
```bash
-filter:v "fade=in:st=IN_POINT:d=DURATION, fade=out:st=OUT_POINT:d=DURATION" \
   -filter:a "afade=in:st=IN_POINT:d=DURATION, afade=out:st=OUT_POINT:d=DURATION"
```
With a note like:
* Where `IN_POINT`, `OUT_POINT`, `DURATION` are the time in seconds (expressed as integers). 

And then maybe your specific example in the notes?

Or am I getting too finicky here 🤔😀

kieranjol:
@kfrn  - Not too finicky - I think the 44.1kHz should be removed to make it more generic.
And i think just using the variables that @kfrn mentions would maybe be the best solution.

I've seen folks get confused by hardcoded values before when trying to scale it to their own use case.

retokromer:
I agree with @kfrn and @kieranjol. On our internal website – of which many parts are publicly accessible – in some of our [recipe](https://avpres.net/FFmpeg/im_H264.html) we have a hard-coded version and the generic version of the commands. We made good experience with. Could this be a solution?

kfrn:
> I've seen folks get confused by hardcoded values before when trying to scale it to their own use case.

This is key, I think!

@DaleLore, did you want to update the PR? I'm also happy to make that change if you'd prefer.

kieranjol (approved):
Thanks for making the changes - looks perfect to me, thank you @DaleLore !!

ablwr:
Thanks @DaleLore and everybody for taking this the extra mile!!

retokromer (approved):

retokromer:
Thank you @DaleLore !

kfrn (approved):
Looks perfect, thanks @DaleLore!

kieranjol:
Yay, thanks @dalelore et al

Issue #429, opened 2020-06-01
Creating a style guide for contributors?


kieranjol:
Hi folks - I was chatting to @DaleLore about this after the Pull Request https://github.com/amiaopensource/ffmprovisr/pull/428 was merged and it got me thinking that we should have some sort of style guide and contributing advice. I don't think that we have much style advice for contributors, and it would be best for all if we could present this in advance rather than letting people know during a code review.
I'm not sure what form this should take, but I looked at a bunch of recipes and picked up on a bunch of trends that I think we'd all mention when doing a code review, so best to have them out front in advance.
Perhaps a basic list of principles would be something like:
* Explanations and examples for recipes should be as generic as possible, even if it involves the use of VARIABLE_NAMES  instead of hardcoded values. This allows users to alter the command and scale it to their own use case. For example :https://amiaopensource.github.io/ffmprovisr/#fade_streams 
* Generally, a recipe should just do one thing or solve one problem. Mixing up multiple ffmpeg commands in one could confuse users and add unintended consequences to ffmpeg commands.
* Add the recipe to https://github.com/amiaopensource/ffmprovisr/blob/gh-pages/recipes.txt
* Recipes involving `-filter_complex` can be some of the most verbose and difficult to understand, so breaking these down as much as possible into their relevant sections is ideal: https://amiaopensource.github.io/ffmprovisr/#append_mp3 
* Some complex recipes may include a gif that show the output, such as  https://amiaopensource.github.io/ffmprovisr/#astats 
* Some recipes require an explanatory section which may not fit gracefully into the recipe itself, for example: https://amiaopensource.github.io/ffmprovisr/#basic-rewrap
* Some recipes may refer to a specifc standard or vocabulary, and it's useful to link to these so that the user can scale the recipe to their use case : https://amiaopensource.github.io/ffmprovisr/#bwf 
* Maybe some HTML-specific stuff that I don't know about?

kfrn:
Good call! Great list, @kieranjol - comprehensive, I can't think of anything to add offhand.

I'd be inclined to add this info in the readme under the Contributing section, and link to it (or repeat the relevant parts) in a PR template. What do you reckon? Happy to set that up.

kieranjol:
I think that's the place for it alright, thank you, that would be great!

ablwr:
This is great!

ablwr:
[...]

PR #432, opened 2020-06-03
swapped out intro-leads for h3

todrobbins:
Before:

Screen Shot 2020-06-02 at 6 16 03 PM

After:

Screen Shot 2020-06-02 at 6 16 13 PM

### DIFF
--- a/css/css.css
+++ b/css/css.css
@@ -113,11 +113,6 @@ h4 {
   font-size: 1.2em;
 }
 
-.intro-lead {
-  font-size: 1.2em;
-  font-weight: bold;
-}
-
 code {
   padding: 2px 4px;
   color: #c7254e;
--- a/css/css.css
+++ b/css/css.css
@@ -85,7 +85,7 @@ a:visited {
 }
 
 .license img {
-  max-width: 10em;
+  max-width: 6em;
   display: inline;
 }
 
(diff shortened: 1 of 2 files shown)


kfrn (commented):
Thank you, Tod! Love the screenshots 

Heh, so this is an existing problem that we were talking about just before (https://github.com/amiaopensource/ffmprovisr/issues/429#issuecomment-637864141), but in terms of semantic HTML, these should really be headers rather than `span`s. (Logically, H3s).

@todrobbins, would you be willing to change the `intro-leads` to H3s and see how that looks?

todrobbins:
Ha! I was thinking the same thing actually. Will do!

kieranjol:
Love the "after" screenshot, so much clearer.

todrobbins:
@kfrn ✅ done

Screen Shot 2020-06-03 at 9 31 14 AM

kfrn (approved):
LGTM! Are we still using the `intro-lead` CSS class at all? If not, feel free to delete ie.

todrobbins:
Good catch!

✅ Removed!

ablwr:
Nice work!

kfrn (approved):
Thanks, @todrobbins!

PR #439, opened 2020-07-17
update H.265/HEVC formula

retokromer:
## Checklist

* [x] I've referred to the [Guidelines for contributing](https://github.com/amiaopensource/ffmprovisr/blob/gh-pages/readme.md#guidelines-for-contributing)

The option `--with-x265` is now enabled on the [standard Homebrew](https://github.com/Homebrew/homebrew-core/blob/master/Formula/ffmpeg.rb) formula, as well as on the alternate ones: [homebrew-ffmpeg](https://github.com/homebrew-ffmpeg/homebrew-ffmpeg/blob/master/Formula/ffmpeg.rb) and [AVpres](https://github.com/AVpres/homebrew-formulae/blob/master/ffmpeg.rb).

### DIFF
--- a/index.html
+++ b/index.html
@@ -432,7 +432,6 @@
       <h5>Transcode to H.265/HEVC</h5>
       <p><code>ffmpeg -i <em>input_file</em> -c:v libx265 -pix_fmt yuv420p -c:a copy <em>output_file</em></code></p>
       <p>This command takes an input file and transcodes it to H.265/HEVC in an .mp4 wrapper, keeping the audio codec the same as in the original file.</p>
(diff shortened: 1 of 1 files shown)


kieranjol:
LGTM. The first part of that line is still true, about compiling ffmpeg
with libx265, but the same can be said about all the x264 recipes so I
approve the deletion.

retokromer:
Indeed, the same can be said for many “options.” I suggest that what is enabled in the official Homebrew formula should not be considered a special situation, because that would confuse many.

ablwr (approved):

kfrn (approved):

PR #460, opened 2022-05-10
join files recipe - remove './' from example, fixes #457

### LINKED ISSUE #457: Joining files together, "safe", and relative filenames
Thank you again for ffmprovisr!

I suggest we change [the "join files together" recipe](https://amiaopensource.github.io/ffmprovisr/index.html#join_files) so the filenames don't have the initial `./` because that seems to read as unsafe to ffmpeg.

Today I consulted this recipe and tried:

`ffmpeg -f concat -i list.txt -c copy gift.mp4`

The contents of `list.txt` had relative filenames, as suggested in the recipe:

~~~
file './1.mp4'
  file './2.mp4'
  file './3.mp4'
  file './4.mp4'
  file './5.mp4'
~~~

But I got the error `Unsafe file name './1.mp4'`. Thank you for the link in that recipe to https://trac.ffmpeg.org/wiki/Concatenate -- that plus [this StackOverflow thread](https://stackoverflow.com/questions/38996925/ffmpeg-concat-unsafe-file-name) led me to add `-safe 0` into the command, and it worked.

I was trying to figure out what I had done wrong, and looked at [the link you'd provided regarding relative vs. absolute paths](https://ffmpeg.org/ffmpeg-formats.html#Options) plus [this Trac ticket](https://trac.ffmpeg.org/ticket/5629). It looks like maybe it prefers NOT to have the `./` at the start of the filename, so I changed the lines in `list.txt` to

~~~
file '1.mp4'
  file '2.mp4'
  file '3.mp4'
  file '4.mp4'
  file '5.mp4'
~~~

and ran `ffmpeg -f concat -i list.txt -c copy gift2.mp4` and it seemed to work fine.

I'm using ffmpeg 4.3.4 on Debian buster.

kieranjol:
This hopefully fixes https://github.com/amiaopensource/ffmprovisr/issues/457 . 
I added a note as well that the file path should not be the absolute path in the textfile. I think that this might be less obvious now that the `./` has been removed in the example.
Edit: I didn't use the term 'relative path' as I was worried that this might lead people to want to use the './'


kfrn:
Cheers @kieranjol! That LGTM as a quick update.

Any chance you'd want to expand at all? @brainwane [put it perfectly](https://github.com/amiaopensource/ffmprovisr/issues/457#issuecomment-1122815952):

> So, it sounds like the recipe should advise people: if possible, run the command from the same directory where the files and the txt file reside. Otherwise you'll have to use `safe` (and here's how).

Otherwise, I could look at this later on.

kfrn (approved):

privatezero:
Looks good to me, although I agree that the little additional clarification suggested by @brainwane and @kfrn - I find that whenever I use concat that I tend to do exactly what is mentioned there so seems kind to users to include that tidbit!

kieranjol:
Desperately hoping that I did this correctly, it's been a while since I made commits within github itself! I totally agree that it should have been added in in the first place!

kieranjol:
Also that intro seemed like the best place to put it?

ablwr (approved):

kfrn:
LGTM @kieranjol!

retokromer (approved):

kieranjol:
Can this be merged?

kieranjol:
Just remembered that I can merge it after the approvals! Thx folks.

PR #491, opened 2026-04-07
housekeeping

retokromer:
## Checklist

* [x] I've referred to the [Guidelines for contributing](https://github.com/amiaopensource/ffmprovisr/blob/gh-pages/readme.md#guidelines-for-contributing)

## Changes

- Remove the `recipes.txt` file, because the `FFCommand Engine` that used it is not maintained since 7 years. In addition, some changes that were made here do not longer allow to generate the list semi-automatically.
- Remove the `ffmprovisr` script (for using `ffmprovisr` off-line), because, on one hand, it no longer works properly everywhere and, on the other hand, the changes that are currently under discussion will need another solution. If this change is accepted, I will also remove the corresponding Homebrew formula.
- Update a few links. (The link to the job advert is dead.)


privatezero (approved):
Looks OK to me!

kfrn:
LGTM!

kfrn (approved):

